PR #2, opened 2021-12-12
Working on Channel & User logic

ThePhilgrim:


mantatail.py
```diff
@@ -126,39 +132,39 @@ def __init__(self, port: int) -> None:
...
-            client_socket, client_address = self.listener_socket.accept()
-            self.user = User(client_address[0])
+            user_socket, user_address = self.listener_socket.accept()
+            user = User(user_address[0], user_socket)
             client_thread = threading.Thread(
-                target=self.recv_loop, args=[client_socket], daemon=True
+                target=self.recv_loop, args=[user], daemon=True
             )
-            self.irc_command_handler = IrcCommandHandler(client_socket)
+            self.irc_command_handler = IrcCommandHandler(user)
```
Akuli, line 146: The command handler shouldn't be saved to `self`, for the same reason that `user` isn't saved to self. You can instantiate the command handler (with a local variable) before starting the thread, and pass `args=[user, irc_command_handler]`

ThePhilgrim:
@Akuli Please review this. Should be ready for merge

mantatail.py
```diff
@@ -7,38 +7,38 @@
...
     def create_user_mask(self):
-        self.user_mask = f"{self.nick}!{self.user_name}@{self.host}"
-        print(self.user_mask)
+        return f"{self.nick}!{self.user_name}@{self.host}"
 
 
 class Channel:
     def __init__(self):
-        pass
+        self.user_list = {}
```
Akuli, line 30: This is a confusing variable name: `user_list` is not a list, but a dict.

mantatail.py
```diff
@@ -69,33 +69,44 @@ def handle_motd(self, user_nick):
...
+                    "<channel name>", message
+                ),
             )
+
+            self.generate_error_reply(no_channel_num, no_channel_info)
+
         else:
-            server.channels[message] = Channel()
-            # print("CHANNELS", server.channels)
+            if not message in self.server.channels.keys():
```
Akuli, line 97:
```suggestion
            if message not in self.server.channels.keys():
```

mantatail.py
```diff
@@ -69,33 +69,44 @@ def handle_motd(self, user_nick):
...
+                self.server.channels[message] = Channel()
+
+            if self.user.nick in self.server.channels[message].user_list.keys():
+                return
+            else:
+                self.server.channels[message].user_list[self.user.nick] = self.user
```
Akuli, line 103: This is currently quite repetitive. You can use `not in`:

```suggestion
            if self.user.nick not in self.server.channels[message].user_list.keys():
                self.server.channels[message].user_list[self.user.nick] = self.user
```

Or just unconditionally add the user, it doesn't matter if it's already added:

```suggestion
            self.server.channels[message].user_list[self.user.nick] = self.user
```

Akuli (commented):
Looks like it should work :) A few possible code cleanups below.

PR #3, opened 2021-12-13
Working on more logic

### LINKED ISSUE #1: 100% cpu usage
1. Run MantaTail
2. Connect a different program (e.g. netcat or Mantaray) to MantaTail
3. Disconnect the program from MantaTail
4. Look at CPU usage. It has jumped to 100% the moment you disconnected.

The problem is here:

https://github.com/ThePhilgrim/MantaTail/blob/9cf4748195bb7eec97335fd7d925a27573fce042/mantatail.py#L53-L54

Once the client has disconnected, `client.socket.recv()` will always return the empty bytes object, so it will keep adding empty bytes objects until MantaTail is stopped with Ctrl+C.

ThePhilgrim:
Not ready for merge.

Fixes #1


mantatail.py
```diff
@@ -109,10 +82,22 @@ def handle_join(self, message):
         #   * Not joined too many channels
 
     def handle_part(self, message):
-        pass
+        if message not in self.server.channels.keys():
+            self.handle_no_such_channel(message)
```
Akuli, line 86: This would be much less confusing if the argument was named e.g. `channel_name` instead of `message`. This also applies to `handle_no_such_channel`.

get_irc_response_args.py
```diff
@@ -0,0 +1,82 @@
...
+        ),
+    )
+
+
+def not_on_channel(message: str) -> Tuple[str, str]:
+    return (
+        irc_response_nums["error_replies"]["ERR_NOTONCHANNEL"][0],
+        irc_response_nums["error_replies"]["ERR_NOTONCHANNEL"][1].replace(
+            "<channel>", message
+        ),
```
Akuli, line 38: - These functions are repetitive. Consider making a single function that does everything they do, and using that instead of copy/pasta.
- `.format()` might be slightly nicer than `.replace()`.

mantatail.py
```diff
@@ -159,36 +166,48 @@ def run_server_forever(self) -> None:
...
+                # IRC messages always end with b"\r\n"
+                while not request.endswith(b"\r\n"):
+                    request_chunk = user.socket.recv(10)
+                    print(request_chunk)
+                    if request_chunk:
+                        request += request_chunk
+                    else:
+                        print(f"{user.nick} has disconnected.")
+                        # user.socket.close()
+                        break
```
Akuli, line 181: If you change this to `return`, you don't need the seemingly unrelated `if not request: break` below.

mantatail.py
```diff
@@ -159,36 +166,48 @@ def run_server_forever(self) -> None:
...
+
+                    verb_lower = verb.lower()
+
+                    if verb_lower == "nick":
+                        user.nick = message
+                        command_handler.handle_motd()
+
+                    if verb_lower not in self.supported_commands:
+                        command_handler.handle_unknown_command(verb_lower)
+                        return
```
Akuli, line 200: The `self.supported_commands` list looks like it will get outdated easily, because it repeats the same info that is already in `IrcCommandHandler`. Here are a couple alternative approaches that don't repeat anything:
- Use `hasattr()` to check if the handler method exists, before trying to find it with `getattr()`
- Put the `getattr()` call inside `try: ... except AttributeError:`.

mantatail.py
```diff
@@ -39,97 +40,102 @@ def __init__(self, server, user):
...
-        motd_num = irc_response_nums["command_responses"]["RPL_MOTD"][0]
-        end_num, end_info = (
-            irc_response_nums["command_responses"]["RPL_ENDOFMOTD"][0],
-            irc_response_nums["command_responses"]["RPL_ENDOFMOTD"][1],
-        )
+        (
+            start_num,
+            start_info,
+        ) = irc_responses.RPL_MOTDSTART
+        motd_num = irc_responses.RPL_MOTD[0]
```
Akuli, line 47: This doesn't seem to use `irc_responses.RPL_MOTD[1]` for anything. Should it be used?

mantatail.py
```diff
@@ -39,97 +40,102 @@ def __init__(self, server, user):
...
         #   * Nick/user not matching bans
         #   * Eventual password matches
         #   * Not joined too many channels
 
-    def handle_part(self, message):
-        pass
+    def handle_part(self, channel_name):
+        if channel_name not in self.server.channels.keys():
+            self.handle_no_such_channel(channel_name)
+        elif self.user.nick not in self.server.channels[channel_name].user_dict.keys():
```
Akuli, line 97: Channels and nicknames should be case insensitive. Feel free to fix this in a separate pull request if you want.

irc_responses.py
```diff
@@ -0,0 +1,15 @@
...
+ERR_NOSUCHSERVER = ("402", ":No such server")
+ERR_NOSUCHCHANNEL = ("403", ":No such channel")
+ERR_CANNOTSENDTOCHAN = ("404", ":Cannot send to channel")
+ERR_TOOMANYCHANNELS = ("405", ":You have joined too many channels")
+ERR_UNKNOWNCOMMAND = ("421", ":Unknown command")
+ERR_NOTONCHANNEL = ("442", ":You're not on that channel")
+
+# Command Responses
+RPL_MOTDSTART = ("375", "Message of the day - ")
+RPL_MOTD = ("372", ":- <text>")
```
Akuli, line 12: Maybe this should use `.format()` too?

Akuli (commented):
Mostly this looks good!

mantatail.py
```diff
@@ -39,97 +33,111 @@ def __init__(self, server, user):
...
-        motd_num = irc_response_nums["command_responses"]["RPL_MOTD"][0]
-        end_num, end_info = (
-            irc_response_nums["command_responses"]["RPL_ENDOFMOTD"][0],
-            irc_response_nums["command_responses"]["RPL_ENDOFMOTD"][1],
-        )
+        (
+            start_num,
+            start_info,
+        ) = irc_responses.RPL_MOTDSTART
+        motd_num = irc_responses.RPL_MOTD[0]
```
Akuli, line 40: Currently this is `"3"`, because `RPL_MOTD` is a string. This is the kind of thing that tests would catch :)

mantatail.py
```diff
@@ -39,97 +33,111 @@ def __init__(self, server, user):
...
+                not_on_channel_num,
+                not_on_channel_info,
+            ) = irc_responses.ERR_NOTONCHANNEL
+
+            self.generate_error_reply(
+                not_on_channel_num, not_on_channel_info, channel_name
+            )
+        else:
+            del self.server.channels[channel_name].user_dict[self.user.nick]
+            if len(self.server.channels[channel_name].user_dict.keys()) == 0:
```
Akuli, line 110: Nit: 
```suggestion
            if len(self.server.channels[channel_name].user_dict) == 0:
```

PR #9, opened 2021-12-16
Start working on tests

ThePhilgrim:
@Akuli I pushed what I have so you have the chance to debug if you get the time.

Todo list (feel free to edit):
- [x] Edit `.github/workflows/github_actions.yml` to also run pytest
- [x] Edit `.github/workflows/github_actions.yml` to run `black` on all python files, `mypy` on all files except tests
- [x] Create a second test. There will be a lot of copy/pasta, use a pytest fixture to clean it up.

PR #10, opened 2021-12-16
prevent duplicates in github actions

Akuli:
When working on #9, we were confused by how pytest had run twice, and we were looking at the output of different pytest runs. This PR makes pytest, mypy and black only run once on each pull request.

I don't know how this configuration works, and I copy/pasted it from somewhere a long time ago. But I know it does work. I have used it a lot on many projects, including Porcupine and Mantaray.

### DIFF
--- a/.github/workflows/github_actions.yml
+++ b/.github/workflows/github_actions.yml
@@ -1,5 +1,9 @@
-# https://medium.com/@doedotdev/mypy-for-github-action-7da1ebee99e7
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
 jobs:
   black:
     runs-on: ubuntu-latest


.github/workflows/github_actions.yml
```diff
@@ -1,5 +1,9 @@
-# https://medium.com/@doedotdev/mypy-for-github-action-7da1ebee99e7
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
```
ThePhilgrim, line 6: The indentation looks a bit weird.  `jobs:` is on the same level as `on:`. In my head, `jobs:` would be indented under `pull_request`
I see that each test only ran once here though..

Akuli: `jobs:` is indented two spaces less than `pull_request:`, so there's nothing inside the `pull_request:` part.

PR #19, opened 2021-12-25
Implement logic to make mantatail work with netcat

### LINKED ISSUE #4: Mantatail is unusable with netcat
It is possible to use libera with just netcat. Run `nc -v irc.libera.chat 6667`, and then type something like this to netcat:

```
NICK akuli4
USER akuli4 0 * :akuli4
```

After a while, you see the motd, and you can then join channels with netcat:

```
JOIN #akuli-testing
PRIVMSG #akuli-testing :hello
```

If I run mantatail and I use `nc -v localhost 6667`, it connects, but typing commands to netcat doesn't do anything.

ThePhilgrim:
Fixes #4

### DIFF
--- a/mantatail.py
+++ b/mantatail.py
@@ -251,7 +251,7 @@ class IrcCommandHandler:
         )
 
 
-def split_on_new_line(string):
+def split_on_new_line(string: str) -> List[str]:
     if string.endswith("\r\n"):
         return string.split("\r\n")
     else:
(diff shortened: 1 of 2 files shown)


mantatail.py
```diff
@@ -251,6 +251,13 @@ def generate_error_reply(
         )
 
 
+def split_on_new_line(string):
```
Akuli, line 254:
```suggestion
def split_on_new_line(string: str) -> List[str]:
```

PR #20, opened 2021-12-26
Cleanup: move user instantiation check inside else statement

Akuli:
We want the user instantiation to happen if `user is None`, and both _nick and _user_name have been given. Having this inside the `else` part of `if user is not None:` makes sense to me, because it contains other related code that runs only when the user is None.

### DIFF
--- a/mantatail.py
+++ b/mantatail.py
@@ -96,12 +96,12 @@ class Server:
                                 )
                             )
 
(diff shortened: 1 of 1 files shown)


ThePhilgrim:
This makes a lot of sense. Thank you.

PR #23, opened 2021-12-26
add pyflakes linter

Akuli:
This is the same tool that Porcupine uses internally for displaying red and yellow underlines in code. It warns about unnecessary imports, potentially undefined variables, and a few other things

I switched the order of the if,else statement in `recv_loop()` to make pyflakes happy. It "reads" the code from top to bottom, and assumes everything is defined before (i.e. above) it is used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,4 +48,4 @@ To make sure the PR will pass these checks, please use the following commands be
 - `black *.py tests/*.py`
 - `mypy --strict *.py`
 - `python3 -m pytest tests/*.py`
-- `python3 -m pyflakes *.py`
+- `python3 -m pyflakes *.py tests/`
--- a/README.md
+++ b/README.md
@@ -41,10 +41,11 @@ Some handy resources for developing IRC-related programs:
 
 ### Pull Requests
 
-Any pull request will automatically be checked for proper format by `black`, `mypy` (--strict mode), as well as by `pytest`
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -41,10 +41,11 @@ Some handy resources for developing IRC-related programs:
...
 
-Any pull request will automatically be checked for proper format by `black`, `mypy` (--strict mode), as well as by `pytest`
+Any pull request will automatically be checked for proper format by `black`, `pyflakes`, `mypy` (--strict mode), as well as by `pytest`
 
 To make sure the PR will pass these checks, please use the following commands before pushing:
 
 - `black *.py tests/*.py`
 - `mypy --strict *.py`
 - `python3 -m pytest tests/*.py`
+- `python3 -m pyflakes *.py`
```
Akuli, line 51:
```suggestion
- `python3 -m pyflakes *.py tests/`
```

ThePhilgrim: @Akuli Doesn't it have to be `python3 -m pyflakes *.py /tests/*.py`?

ThePhilgrim:
@Akuli I'm on IRC if you wanna come early today

README.md
Akuli: - Definitely not `/tests/`, as that is an absolute path that likely doesn't exist, and outside the `mantatail` repo.
- It accepts folders or files, so `tests/*.py` would work too.

PR #24, opened 2021-12-26
Add separate tests for different platforms

### LINKED ISSUE #11: Running tests on multiple platforms
Ideally github actions would run the tests on windows, macos and linux. This way, if I write a pull request that won't work on macos, ThePhilgrim will notice the problem before it gets merged.

Many repositories already have github actions config for this that we can copy, e.g. Akuli/porcupine, Akuli/mantaray, python/typeshed to name a few that I have worked on recently.

ThePhilgrim:
Fixes #11

### DIFF
--- a/.github/workflows/github_actions.yml
+++ b/.github/workflows/github_actions.yml
@@ -38,46 +38,15 @@ jobs:
           python3 -m pip install pyflakes
       - run: |
           pyflakes *.py tests
-  pytest-windows:
+  pytest:
     timeout-minutes: 2
-    strategy:
-      matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10']
-    env:
-      TCLLIBPATH: lib
-    runs-on: windows-latest
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
-        with:
-          python-version: ${{ matrix.python-version }}
-      - run: pip install -r requirements.txt -r requirements-dev.txt
-      - run: python -m pytest
-  pytest-linux:
-    timeout-minutes: 2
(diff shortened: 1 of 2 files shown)


.github/workflows/github_actions.yml
```diff
@@ -38,15 +38,33 @@ jobs:
...
+      matrix:
+        # os: ["ubuntu-latest", "windows-latest", "macos-latest"]
+        os: ['ubuntu-latest', 'windows-latest']
+        python-version: ['3.6', '3.7', '3.8', '3.9', '3.10']
     steps:
-      - uses: actions/checkout@v1
-      - uses: actions/setup-python@v1
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }} on ${{ matrix.os }}
+      - uses: actions/setup-python@v2
```
Akuli, line 64: Typo:
```suggestion
        uses: actions/setup-python@v2
```

Akuli: No longer needed, you changed a different thing.

.github/workflows/github_actions.yml
```diff
@@ -40,13 +40,17 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
         with:
-          python-version: '3.9'
+          python-version: ${{ matrix.python-version }}
       - run: |
           python3 -m pip install pytest
       - run: |
-          python3 -m pytest tests/*.py
+          python3 -m pytest
```
Akuli, line 56: You might want to update README too, since it contains the same command.

Akuli (approved):

PR #25, opened 2021-12-26
Work on sending back msg upon JOIN

ThePhilgrim:


mantatail.py
```diff
@@ -71,30 +70,21 @@ def recv_loop(self, user_info: Tuple[str, socket.socket]) -> None:
 
                     if user is None:
                         if verb_lower == "user":
-                            _user_name = message.split(" ", 1)[0]
+                            _user_name = message
```
Akuli, line 73: I don't think this is correct: https://modern.ircdocs.horse/#user-message

If Bob sends `USER BobUsr 0 * :Bob's real name`, the username should be set to `BobUsr`.

mantatail.py
```diff
@@ -71,30 +70,21 @@ def recv_loop(self, user_info: Tuple[str, socket.socket]) -> None:
...
-                            )
+                            (error_code, error_info) = irc_responses.ERR_NOTREGISTERED
+                            user_socket.sendall(bytes(f":mantatail {error_code} * {error_info}\r\n", encoding="utf-8"))
 
                         if _user_name and _nick:
                             user = User(user_host, user_socket, _user_name, _nick)
-                            command_handler: IrcCommandHandler = IrcCommandHandler(
-                                self, user
-                            )
+                            command_handler: IrcCommandHandler = IrcCommandHandler(self, user)
```
Akuli, line 82: Should work like this, after the code change in #23:

```suggestion
                            command_handler = IrcCommandHandler(self, user)
```

mantatail.py
```diff
@@ -181,40 +168,29 @@ def handle_join(self, channel_name: str) -> None:
...
+            if lower_user_nick not in self.server.channels[lower_channel_name].user_dict.keys():
+                self.server.channels[lower_channel_name].user_dict[lower_user_nick] = self.user
+                self.user.socket.sendall(bytes(f":{self.user.user_mask} JOIN {channel_name}{self.send_to_client_suffix}", encoding=self.encoding))
 
         # TODO: Check for:
         #   * User invited to channel
         #   * Nick/user not matching bans
         #   * Eventual password matches
         #   * Not joined too many channels
+        #   * Forward to another channel (irc num 470) ex. #homebrew -> ##homebrew
```
Akuli, line 183: This probably shouldn't be in a "Check for:" list, because a server can just decide to not do any forwarding. But setting up forwarding would be useful for developing mantaray.

tests/test_mantatail.py
```diff
@@ -66,7 +66,7 @@ def user_alice(run_server):
     alice_socket = socket.socket()
     alice_socket.connect(("localhost", 6667))
     alice_socket.sendall(b"NICK Alice\r\n")
-    alice_socket.sendall(b"USER AliceUsr 0 * :Alice's real name\r\n")
+    alice_socket.sendall(b"USER Alice 0 * :Alice's real name\r\n")
```
Akuli, line 69: It would be good to distinguish between the user and nick, so that you will notice if you accidentally use the wrong one.

tests/test_mantatail.py
```diff
@@ -123,6 +123,18 @@ def test_join_before_registering(run_server):
     assert received == b":mantatail 451 * :You have not registered\r\n"
 
 
+def test_join_channel(user_alice, user_bob):
+    user_alice.sendall(b"JOIN #foo\r\n")
+    user_bob.sendall(b"JOIN #foo\r\n")
+    received = receive_line(user_bob)
+    print(received)
+    assert received == b":Bob!Bob 0 * :Bob's real name@127.0.0.1 JOIN #foo\r\n"
```
Akuli, line 131: This isn't correct. IRC clients should be able to just assume that they can `.split(" ")`, but this won't work if you send the real name and it contains spaces.

To see what libera does, you can set a different real name in mantaray (right-click irc.libera.chat, "Server settings").

```suggestion
    assert received == b":Bob!BobUsr@127.0.0.1 JOIN #foo\r\n"
```

ThePhilgrim:
Currently, if I run pytest many times, it fails maybe 40% of the times. The reason is the code below. The test can be fixed with the commented out time.sleep(1). **However**, there might be an actual error if two users join a channel at literally the same time. The risk might be low, but it's there. I would need some help with figuring out how to protect the code against this. 

```
for user in channel_user_keys:
                    self.server.channels[lower_channel_name].user_dict[
                        user
                    ].socket.sendall(
                        bytes(
                            f":{self.user.user_mask} JOIN {channel_name}{self.send_to_client_suffix}",
                            encoding=self.encoding,
                        )
                    )
```

Akuli:
Indeed, things can go wrong if two users join at exactly the same time. This is known as a **race condition**.

The usual solution to this is a lock. In code, it looks like this:

```
lock = threading.Lock()

def thread1():
    with lock:
        do stuff

def thread2():
    with lock:
        do stuff
```

Only one thread at a time can execute code under `with lock:`. If thread 1 is already executing it when thread 2 wants to run it, it will wait until thread 1 has finished doing whatever it does in `with lock:`.

Here's a minimal example. If you run it, you will see that only one thread executes a `time.sleep(1)` at a time. In your code, you need `with lock:` around anything that e.g. assumes that other threads are not changing a dictionary.

```py
import threading
import time

lock = threading.Lock()

def thread1():
    with lock:
        print("Thread 1 doing stuff")
        time.sleep(1)
        print("Thread 1 DONE")

def thread2():
    with lock:
        print("Thread 2 doing stuff")
        time.sleep(2)
        print("Thread 2 DONE")

threading.Thread(target=thread1).start()
threading.Thread(target=thread2).start()
```

mantatail.py
```diff
@@ -10,9 +10,11 @@
 
 class Server:
     def __init__(self, port: int, motd_content: Optional[Dict[str, List[str]]]) -> None:
+        print("Starting Mantatail...")
         self.host = "127.0.0.1"
         self.port = port
         self.motd_content = motd_content
+        self.thread_lock = threading.Lock()
```
Akuli, line 17: Consider naming the lock so that it is clear when you have to hold it. E.g. `self.channels_and_users_lock`

mantatail.py
```diff
@@ -51,15 +54,17 @@ def recv_loop(self, user_info: Tuple[str, socket.socket]) -> None:
...
                             print("Disconnected.")
                         return
 
                 decoded_message = request.decode("utf-8")
                 for line in split_on_new_line(decoded_message)[:-1]:
                     if " " in line:
-                        verb, message = line.split(" ", 1)
+                        (
+                            verb,
+                            message,
```
Akuli, line 66: If you delete the `,` after `message`, black will format this sanely. You can also add `skip_magic_trailing_comma = true` in the `[tool.black]` section of `pyproject.toml`, to make black not assign a magic meaning to trailing commas.

mantatail.py
```diff
@@ -207,10 +258,7 @@ def handle_part(self, channel_name: str) -> None:
             lower_user_nick
             not in self.server.channels[lower_channel_name].user_dict.keys()
```
Akuli, line 259: These checks look like they should be done while holding the lock.

PR #27, opened 2021-12-26
Implement logic to handle closing user connection

ThePhilgrim:


Akuli:
Needs a test.

Akuli:
...or not, because the tests already send `QUIT`

mantatail.py
```diff
@@ -219,8 +222,9 @@ def handle_part(self, channel_name: str) -> None:
 
         # TODO: Support user writing /part without specifying channel name
```
Akuli, line 223: I don't think this is something a server has to worry about. The server doesn't know which of the joined channels the user is currently looking at.

PR #30, opened 2021-12-28
Add thread lock for part channel

### LINKED ISSUE #29: join/part race condition
In the following test, Alice and Bob both join a channel, randomly wait 0 to 10 milliseconds, and then part the channel. This is repeated 100 times.

```py
def test_join_part_race_condition(user_alice, user_bob):
    for i in range(100):
        user_alice.sendall(b"JOIN #foo\r\n")
        time.sleep(random.randint(0, 10) / 1000)
        user_alice.sendall(b"PART #foo\r\n")

        user_bob.sendall(b"JOIN #foo\r\n")
        time.sleep(random.randint(0, 10) / 1000)
        user_bob.sendall(b"PART #foo\r\n")
```

The test fails on main.

ThePhilgrim:
Fixes #29

### DIFF
--- a/mantatail.py
+++ b/mantatail.py
@@ -247,21 +247,25 @@ class IrcCommandHandler:
     def handle_part(self, channel_name: str) -> None:
         lower_channel_name = channel_name.lower()
         lower_user_nick = self.user.nick.lower()
-        if lower_channel_name not in self.server.channels.keys():
-            self.handle_no_such_channel(channel_name)
-        elif (
-            lower_user_nick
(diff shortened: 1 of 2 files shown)


ThePhilgrim:
@Akuli, I would prefer if you pushed the test in issue #29 to this PR, since you wrote the code. Right should be right!
Your test passes with the pushed code.

ThePhilgrim:
@Akuli If you are available on IRC, I'd really need your help with a mantatail test that I've been trying to debug for hours.

PR #31, opened 2021-12-29
Implement logic to PRIVMSG a channel

ThePhilgrim:


tests/test_mantatail.py
```diff
@@ -152,6 +152,24 @@ def test_youre_not_on_that_channel(user_alice, user_bob):
...
+    while receive_line(user_alice) != b":Bob!BobUsr@127.0.0.1 PRIVMSG #foo :Foo\r\n":
+        pass
+
+    user_alice.sendall(b"PRIVMSG #foo :Bar\r\n")
+    while receive_line(user_bob) != b":Alice!AliceUsr@127.0.0.1 PRIVMSG #foo :Bar\r\n":
+        pass
+
+    user_bob.sendall(b"PRIVMSG #foo :Baz\r\n")
+    while receive_line(user_alice) != b":Bob!BobUsr@127.0.0.1 PRIVMSG #foo :Baz\r\n":
+        pass
```
Akuli, line 170: Should also test what errors if you get, if you try to send a message to a channel you haven't joined, or a channel that doesn't exist.

Also, does this test really need while loops? The purpose of while loops in other tests is to skip over several uninteresting lines, when we want to ensure that last line matches our expectations.

mantatail.py
```diff
@@ -245,6 +245,7 @@ def handle_join(self, channel_name: str) -> None:
         #   * Forward to another channel (irc num 470) ex. #homebrew -> ##homebrew
 
     def handle_part(self, channel_name: str) -> None:
+        # TODO: Show part message to other users & Remove from user list.
```
Akuli, line 248: "Remove from user list" sounds like QUIT to me, not PART.

mantatail.py
```diff
@@ -270,7 +271,35 @@ def handle_quit(self, message: str) -> None:
...
+            lower_channel_name = receiver.lower()
+            channel_user_keys = self.server.channels[
+                lower_channel_name
+            ].user_dict.keys()
+
+            sender_user_mask = (
+                self.server.channels[lower_channel_name]
+                .user_dict[self.user.nick.lower()]
+                .user_mask
+            )
```
Akuli, line 288: This should probably be done while holding the lock. You don't want the channel to disappear, or users to join/part, while you are trying to use its `user_dict`.

tests/test_mantatail.py
ThePhilgrim: It seems like the while loops are needed here, since the server sends back join messages, channel users, end of user list, etc.

There are several messages that are sent before the actual message.

mantatail.py
ThePhilgrim: This means that the user has to be removed from the channel.user_list. So in other words, the users that are on a specific channel

mantatail.py
ThePhilgrim: That makes a lot of sense. Will fix this.

tests/test_mantatail.py
Akuli: Everything you mentioned seems to have to do with joining `#foo`, so two while loops after `JOIN #foo`, one for each user, should be enough.

ThePhilgrim: Are you saying that it would be better to first have two while loops that passes through the join messages after `JOIN #foo` and then an assert after the `PRIVMSG`?

That would be more code. Any reason you think this is better?

mantatail.py
```diff
@@ -117,7 +117,7 @@ def __init__(self, channel_name: str, channel_creator: str) -> None:
         self.name = channel_name
         self.creator = channel_creator
         self.topic = None
-        self.user_dict: Dict[Optional[str], User] = {}
+        self.user_dict: Dict[Optional[str], User,] = {}
```
Akuli, line 120: There's several changes like this. Do you like them?

mantatail.py
```diff
@@ -250,7 +252,35 @@ def handle_quit(self, message: str) -> None:
     def _handle_kick(self, message: str) -> None:
         pass
 
-    def _handle_privmsg(self, message: str) -> None:
+    def handle_privmsg(self, message: str) -> None:
+        with self.server.channels_and_users_thread_lock:
+            (receiver, privmsg) = message.split(" ", 1)
```
Akuli, line 257: Does `privmsg` start with a `:` character? It's fine if it does, but the variable name should be `colon_privmsg` or something, to make it clear that there's more to it than just the message. It would be also good to `assert` that it indeed starts with `:`, if you expect it to start with `:`.

mantatail.py
```diff
@@ -250,7 +252,35 @@ def handle_quit(self, message: str) -> None:
...
+                cant_send_num, cant_send_info = irc_responses.ERR_CANNOTSENDTOCHAN
+                self.generate_error_reply(cant_send_num, cant_send_info, receiver)
+            else:
+                sender_user_mask = (
+                    self.server.channels[lower_channel_name].user_dict[lower_sender_nick].user_mask
+                )
+
+                for user in self.server.channels[lower_channel_name].user_dict.keys():
+                    if user != lower_sender_nick:
+                        self.server.channels[lower_channel_name].user_dict[user].socket.sendall(
```
Akuli, line 276: `.items()` would make this simpler.

tests/test_mantatail.py
```diff
@@ -151,6 +151,39 @@ def test_youre_not_on_that_channel(user_alice, user_bob):
...
+
+    user_bob.sendall(b"PRIVMSG #foo :Foo\r\n")
+    assert receive_line(user_alice) == b":Bob!BobUsr@127.0.0.1 PRIVMSG #foo :Foo\r\n"
+
+    user_alice.sendall(b"PRIVMSG #foo :Bar\r\n")
+    assert receive_line(user_bob) == b":Alice!AliceUsr@127.0.0.1 PRIVMSG #foo :Bar\r\n"
+
+    user_bob.sendall(b"PRIVMSG #foo :Baz\r\n")
+    assert receive_line(user_alice) == b":Bob!BobUsr@127.0.0.1 PRIVMSG #foo :Baz\r\n"
```
Akuli, line 172: Sending multiple words should be tested too, because accidentally writing `.split(" ", 2)` would be enough to break it:

```suggestion
    user_bob.sendall(b"PRIVMSG #foo :Hello world\r\n")
    assert receive_line(user_alice) == b":Bob!BobUsr@127.0.0.1 PRIVMSG #foo :Hello world\r\n"
```

PR #32, opened 2021-12-30
Refactor IrcCommandHandler into separate file

ThePhilgrim:
Pytest was complaining about circular imports. I tried to use if TYPE_CHECKING, but with no luck. Pytest gave different AttributeErrors, for example, it didn't like `func(server: Server)`.

I got some help in `#python`, and the best solution we could get to was importing handle inside of `recv_loop()`.
They did however say that imports in method usually means smelly code. He thought the best option was to put the functions in `handle.py` in a class.

He also said he didn't like the `getattr()` solution for a module. He talked about a `@register` decorator to use with a dict. I'm not sure how that would be done, if it's really a better solution. What do you think?

Here are parts of our conversation:

Skärmavbild 2021-12-30 kl  01 31 12

Skärmavbild 2021-12-30 kl  01 31 39

### DIFF
--- a/handle.py
+++ b/handle.py
@@ -103,7 +103,7 @@ def part(server: mantatail.Server, user: mantatail.User, channel_name: str) -> N
 
 
 # !Not implemented
-def _kick(self, message: str) -> None:
+def _kick(message: str) -> None:
     pass
 
 
@@ -113,7 +113,6 @@ def quit(server: mantatail.Server, user: mantatail.User, channel_name: str) -> N
 
 
 def privmsg(server: mantatail.Server, user: mantatail.User, msg: str) -> None:
-    print("NEW")
     with server.channels_and_users_thread_lock:
         (receiver, colon_privmsg) = msg.split(" ", 1)
 
@@ -141,7 +140,7 @@ def privmsg(server: mantatail.Server, user: mantatail.User, msg: str) -> None:
 
 
 # !Not implemented
-def privmsg_to_user(receiver, colon_privmsg) -> None:
+def privmsg_to_user(receiver: str, colon_privmsg: str) -> None:
     pass
 
 
@@ -154,6 +153,13 @@ def error_unknown_command(user: mantatail.User, command: str) -> None:
     send_bytes_to_user(user, error_message_as_bytes)
 
 
+def error_not_registered() -> bytes:
+    (not_registered_num, not_registered_info) = irc_responses.ERR_NOTREGISTERED
+
+    message = f"{not_registered_num} * {not_registered_info}"
+    return convert_string_to_server_message(message)
+
+
 def error_no_motd(user: mantatail.User) -> None:
     (no_motd_num, no_motd_info) = irc_responses.ERR_NOMOTD
 
@@ -162,7 +168,7 @@ def error_no_motd(user: mantatail.User) -> None:
     send_bytes_to_user(user, error_message_as_bytes)
 
 
-def error_no_such_nick_channel(user: mantatail.User, channel_name) -> None:
+def error_no_such_nick_channel(user: mantatail.User, channel_name: str) -> None:
     (no_nick_num, no_nick_info) = irc_responses.ERR_NOSUCHNICK
 
     message = f"{no_nick_num} {channel_name} {no_nick_info}"
@@ -178,7 +184,7 @@ def error_not_on_channel(user: mantatail.User, channel_name: str) -> None:
     send_bytes_to_user(user, error_message_as_bytes)
 
 
-def error_cannot_send_to_channel(user: mantatail.User, channel_name) -> None:
+def error_cannot_send_to_channel(user: mantatail.User, channel_name: str) -> None:
     (cant_send_num, cant_send_info) = irc_responses.ERR_CANNOTSENDTOCHAN
 
     message = f"{cant_send_num} {channel_name} {cant_send_info}"
@@ -195,7 +201,7 @@ def error_no_such_channel(user: mantatail.User, channel_name: str) -> None:
 
 
 ### Actions
-def convert_string_to_server_message(message: str, prefix="mantatail") -> bytes:
+def convert_string_to_server_message(message: str, prefix: str = "mantatail") -> bytes:
     utf_8 = "utf-8"
     suffix = "\r\n"
     return bytes(f":{prefix} {message}{suffix}", encoding=utf_8)
--- a/handle.py
+++ b/handle.py
@@ -33,7 +33,9 @@ def motd(server: mantatail.Server, user: mantatail.User) -> None:
 
 
 def join(server: mantatail.Server, user: mantatail.User, channel_name: str) -> None:
-    channel_regex = r"#[^ \x07,]{1,49}"  # TODO: Make more restrictive (currently valid: ###, #ö?!~ etc)
+    channel_regex = (
+        r"#[^ \x07,]{1,49}"  # TODO: Make more restrictive (currently valid: ###, #ö?!~ etc)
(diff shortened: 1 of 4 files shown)


mantatail.py
```diff
@@ -66,35 +64,26 @@ def recv_loop(self, user_info: Tuple[str, socket.socket]) -> None:
...
+                            handle.motd(self, user)
 
                     else:
                         try:
-                            call_handler_function = getattr(
-                                command_handler, handler_function_to_call
-                            )
-                            call_handler_function(message)
+                            # ex. "handle.nick" or "handle.join"
+                            call_handler_function = getattr(handle, verb_lower)
```
Akuli, line 82: As mentioned in the screenshots, it is currently possible for users to call any function defined in `handle.py`, which isn't secure if you define or import a function that deletes a file, for example. It would be better to have some prefix in the function names, e.g. `handle_join()` instead of `join()`. This way, only the functions whose name starts with `handle_` can be called by sending a message to the server.

handle.py
```diff
@@ -0,0 +1,218 @@
...
+
+    message = f"{no_channel_num} {channel_name} {no_channel_info}"
+    error_message_as_bytes = convert_string_to_server_message(message)
+    send_bytes_to_user(user, error_message_as_bytes)
+
+
+### Actions
+def convert_string_to_server_message(message: str, prefix: str = "mantatail") -> bytes:
+    utf_8 = "utf-8"
+    suffix = "\r\n"
```
Akuli, line 213: What is the purpose of these variables?

handle.py
```diff
@@ -0,0 +1,218 @@
...
+    error_message_as_bytes = convert_string_to_server_message(message)
+    send_bytes_to_user(user, error_message_as_bytes)
+
+
+def error_no_such_nick_channel(user: mantatail.User, channel_name: str) -> None:
+    (no_nick_num, no_nick_info) = irc_responses.ERR_NOSUCHNICK
+
+    message = f"{no_nick_num} {channel_name} {no_nick_info}"
+    error_message_as_bytes = convert_string_to_server_message(message)
+    send_bytes_to_user(user, error_message_as_bytes)
```
Akuli, line 183: These two lines are repeated a lot. It might be cleaner if `send_bytes_to_user()` would internally call `convert_string_to_server_message()`, and be renamed accordingly, e.g. `send_string_to_user()`.

Also, why is `send_bytes_to_user` not a method on `class User`, since it needs a User object anyway?

Akuli:
Regarding a `@register` decorator and a dict of functions: Mantaray had a `@register` decorator (named differently), and I deleted it because it was more complicated than it needed to be. The exactly same thing happened in Porcupine's pastebin plugin. See `_define_commands()` in mantaray's current `commands.py` for what ChrisWarrick had in mind.

Akuli:
Circular imports: The problem was that when `mantatail.py` was imported, it imported `handle` before defining its classes, and then `handle` imported the not-yet-loaded `mantatail` and tried to do `mantatail.Server` in a type annotation. If the import is moved inside a function, it runs much later, when both files are fully loaded and they contain everything you defined in them.

An easy and not smelly fix is to add `from __future__ import annotations` to the beginning of `handle.py`. This causes Python to not evaluate the type annotations at all, so even if they do `mantatail.Server` and `mantatail` isn't fully loaded yet, Python won't notice it. (This is the default behaviour in Python 3.11 and newer.)

tests/test_mantatail.py
```diff
@@ -4,7 +4,9 @@
 import traceback
 import threading
 import time
-from mantatail import Server
+import mantatail
+
+# from mantatail import Server
```
Akuli, line 9: This shouldn't make any difference in `test_mantatail.py`, because both imports will load both `mantatail.py` and `command.py`. But it should help in `command.py`.

mantatail.py
```diff
@@ -65,36 +63,28 @@ def recv_loop(self, user_info: Tuple[str, socket.socket]) -> None:
...
-                                command_handler, handler_function_to_call
-                            )
-                            call_handler_function(message)
+                            # ex. "command.handle_nick" or "command.handle_join"
+                            call_handler_function = getattr(command, parsed_command)
+                            call_handler_function(self, user, message)
+
                         except AttributeError:
-                            command_handler.handle_unknown_command(verb_lower)
+                            command.error_unknown_command(user, verb_lower)
```
Akuli, line 87: To prevent catching AttributeErrors from calling the handler function:

```python
except AttributeError:
    command.error_unknown_command(user, verb_lower)
else:
    call_handler_function(self, user, message)
```

(I can't use GitHub's suggestion button, because it only works when I have not selected any deleted (red) lines)

mantatail.py
```diff
@@ -101,6 +101,16 @@ def __init__(self, host: str, socket: socket.socket, user_message: str, nick: st
         self.user_mask = f"{self.nick}!{self.user_name}@{self.host}"
         self.closed_connection = False
 
+    def send_string_to_user(self, message: str) -> None:
+        message_as_bytes = bytes(f":mantatail {message}\r\n", encoding="utf-8")
+        self.socket.sendall(message_as_bytes)
+
+    def send_string_with_user_prefix(self, message: str, prefix: str) -> None:
+        if not prefix:
+            prefix = self.user_mask
```
Akuli, line 110: When does this run?

mantatail.py
```diff
@@ -101,6 +101,16 @@ def __init__(self, host: str, socket: socket.socket, user_message: str, nick: st
         self.user_mask = f"{self.nick}!{self.user_name}@{self.host}"
         self.closed_connection = False
 
+    def send_string_to_user(self, message: str) -> None:
```
Akuli, line 104: Nit: Because the method is in `class User`, I would name it `send_string()` instead.

mantatail.py
ThePhilgrim: This was me trying to figure out a test that didn't run properly after refactoring the `send_string_to_user()` stuff. The test expected `Alice!AliceUsr@...` but got `Bob!BobUsr@...`. 

Thought process:
- Sometimes the prefix when setting something to a user is `:mantatail`, and sometimes it's the `user_mask`
- I had only one method in `User` for sending string to user. The user would get passed into the `handle_X`from `Server`. 
- `send_string_to_user()` had to know what prefix to use, so it had to take a prefix argument. However, in my mind, it didn't make sense to pass `user.user_mask` as an argument to a method in `User`.
- Therefore I split it up to have a separate method for when user_mask is the prefix. However, I realize that when sending something like `user_mask JOIN #foo\r\n` to Alice, the `user_mask` is Alice's (the receiver), but when sending to the other people in the channel, the user mask is still Alice (the other part), so I figured that I needed for deciding if the user mask should be for the receiver or the user that made an action. I wrote the code at a friends house so I couldn't concentrate as well as I do at home. I never reached the point that I was fully sure this is how the logic should work.

mantatail.py
ThePhilgrim: To be completely honest I'm not sure I like having the `send_string_to_user` in `User`. It feels like it should be there if `User` was the sender, but not if `User` is the receiver as in this case. 

But I guess it makes sense to keep it there anyways. But then `send_string()` makes it even less clear that the `User` is receiving the message and not sending. What are your thoughts?

mantatail.py
ThePhilgrim: After looking at the code for a bit and reading through some IRC docs I realize that it's always the sender's user mask that is the prefix, never the user_mask of the receiver. That means this if check is not needed. Will update accordingly.

Akuli (approved):
Looks reasonable.

PR #33, opened 2021-12-31
Combine send_string_user_mask_prefix and send_string

Akuli:
`send_string` does everything that `send_string_user_mask_prefix` does, but with `"mantatail"` instead of a user's mask as the prefix. Combine both functions together, using a default value of `"mantatail"` for the mask argument.


ThePhilgrim:
@Akuli I'm on irc for a few hours if you want to do some coding early today :)

PR #34, opened 2021-12-31
split Server into Listener and ServerState

Akuli:
We aren't sure yet whether any of this is a good idea.

### DIFF
--- a/command.py
+++ b/command.py
@@ -3,38 +3,40 @@ import re
 import mantatail
 import irc_responses
 
+from typing import Optional, Dict, List
+
 
 ### Handlers
-def handle_join(server: mantatail.Server, user: mantatail.User, channel_name: str) -> None:
+def handle_join(state: mantatail.ServerState, user: mantatail.User, channel_name: str) -> None:
     channel_regex = (
         r"#[^ \x07,]{1,49}"  # TODO: Make more restrictive (currently valid: ###, #ö?!~ etc)
     )
 
     lower_channel_name = channel_name.lower()
-    with server.channels_and_users_thread_lock:
+    with state.lock:
         if not re.match(channel_regex, lower_channel_name):
             error_no_such_channel(user, channel_name)
         else:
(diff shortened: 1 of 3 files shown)


mantatail.py
```diff
@@ -2,92 +2,93 @@
...
+                    else:
+                        call_handler_function(state, user, message)
 
-                        if user.closed_connection:
-                            return
+                    if user.closed_connection:
+                        return
 
 
 class User:
```
ThePhilgrim, line 94:
```suggestion
class UserConnection:
```

I think it makes sense to rename this in this refactor to clarify its responsibility

mantatail.py
```diff
@@ -2,92 +2,93 @@
...
-                            # ex. "command.handle_nick" or "command.handle_join"
-                            call_handler_function = getattr(command, parsed_command)
-                        except AttributeError:
-                            command.error_unknown_command(user, verb_lower)
-                        else:
-                            call_handler_function(self, user, message)
+                        user_socket.sendall(command.error_not_registered())
+
+                    if _user_message and _nick:
+                        user = User(user_host, user_socket, _user_message, _nick)
```
ThePhilgrim, line 78:
```suggestion
                        user = UserConnection(user_host, user_socket, _user_message, _nick)
```

ThePhilgrim:
Can we update `send_string()` to `send_string_to_client()` or do you think that's a bad idea?

ThePhilgrim (changes requested):

command.py
```diff
@@ -52,7 +56,7 @@ def handle_join(server: mantatail.Server, user: mantatail.User, channel_name: st
         #   * Send End of Name list (366)
 
         # TODO: Check for:
-        #   * User invited to channel
+        #   * UserConnection invited to channel
```
Akuli, line 59:
```suggestion
        #   * User invited to channel
```

ThePhilgrim (approved):

PR #35, opened 2021-12-31
Part quit msg

ThePhilgrim:


command.py
```diff
@@ -88,11 +88,18 @@ def _handle_kick(message: str) -> None:
...
-    user.socket.close()
+def handle_quit(state: mantatail.ServerState, user: mantatail.UserConnection, channel_name: str) -> None:
+    # TODO: Implement logic for different reasons & disconnects.
+    reason = "(Remote host closed the connection)"
+
+    for channel_name, channel in state.channels.items():
+        message = f"QUIT {channel_name} Quit: {reason}"
+        if user.nick.lower() in channel.user_dict.keys():
+            for nick, receiver in channel.user_dict.items():
+                if nick != user.nick.lower():
```
Akuli, line 99: I think mantaray expects the server to send a QUIT message to the person who quits. Not getting it might be causing some issues.

command.py
```diff
@@ -88,11 +88,18 @@ def _handle_kick(message: str) -> None:
...
-    user.closed_connection = True
-    user.socket.close()
+def handle_quit(state: mantatail.ServerState, user: mantatail.UserConnection, channel_name: str) -> None:
+    # TODO: Implement logic for different reasons & disconnects.
+    reason = "(Remote host closed the connection)"
+
+    for channel_name, channel in state.channels.items():
+        message = f"QUIT {channel_name} Quit: {reason}"
+        if user.nick.lower() in channel.user_dict.keys():
+            for nick, receiver in channel.user_dict.items():
```
Akuli, line 98: If a user is on 5 channels, he will receive the QUIT message 5 times. This isn't good.

command.py
ThePhilgrim: @Akuli Does mantaray expect the server to send the QUIT message to the person who quits once, or once per channel though?

Akuli: Once. When it receives a QUIT, it removes that user from all channels where they were.

In code, this is more complicated than it should be. First, every view defines *relevant nicks*, and gets notified for their QUITs:

```py
            elif isinstance(event, backend.UserQuit):
                for view in self.get_subviews(include_server=True):
                    if event.nick in view.get_relevant_nicks():
                        view.on_relevant_user_quit(event.nick, event.reason)
```

Then `ChannelView` defines its relevant nicks to be what is in the userlist, and removes from the userlist when someone quits:

```py
    def get_relevant_nicks(self) -> tuple[str, ...]:
        return self.userlist.get_nicks()

    def on_relevant_user_quit(self, nick: str, reason: str | None) -> None:
        super().on_relevant_user_quit(nick, reason)
        self.userlist.remove_user(nick)
```

command.py
ThePhilgrim: Wouldn't a user get another user's quit message in every channel that they are both in?

I think we need to design the quit functionality together. I'm not exactly sure how this would best be implemented. I will be on IRC most of the day tomorrow if you want to come early.

Akuli: A quit message doesn't have anything to do with channels, and shouldn't even include the channel name. See https://modern.ircdocs.horse/#quit-message

command.py
```diff
@@ -91,8 +99,24 @@ def _handle_kick(message: str) -> None:
 def handle_quit(
     state: mantatail.ServerState, user: mantatail.UserConnection, channel_name: str
```
Akuli, line 100: What does `channel_name` do here?

tests/test_mantatail.py
```diff
@@ -82,6 +82,8 @@ def user_alice(run_server):
 
     yield alice_socket
     alice_socket.sendall(b"QUIT\r\n")
+    while not b"QUIT" in receive_line(alice_socket):
```
Akuli, line 85: Nit: `not foo in bar` is more readable as `foo not in bar`.

PR #37, opened 2022-01-01
Logic for channel operators

ThePhilgrim:


ThePhilgrim:
Proper tests not implemented yet. Currently working on them to add to PR

command.py
```diff
@@ -173,6 +184,60 @@ def motd(motd_content: Optional[Dict[str, List[str]]], user: mantatail.UserConne
     user.send_string_to_client(motd_start_and_end["end_msg"])
 
 
+def process_channel_modes(
+    state: mantatail.ServerState, user: mantatail.UserConnection, args: List[str]
+) -> None:
+    supported_channel_modes = ["o"]
+    if args[0].lower() not in state.channels.keys():
```
Akuli, line 191: This should probably hold `state.lock`. Also, a variable like `channel_name = args[0]` would make things more readable.

command.py
```diff
@@ -173,6 +184,60 @@ def motd(motd_content: Optional[Dict[str, List[str]]], user: mantatail.UserConne
...
+        error_no_such_channel(user, args[0])
+    elif len(args) == 1:
+
+        message = f'{irc_responses.RPL_CHANNELMODEIS} {args[0]} {" ".join(state.channels[args[0].lower()].modes)}'
+        user.send_string_to_client(message)
+    else:
+        assert len(args) == 3
+        target_chan = args[0]
+        mode_command = list(args[1])
+        target_user = args[2]
```
Akuli, line 202: Nit: there is a handy way to do all this.

```suggestion
        target_chan, mode_command, target_user = args
```

This makes `mode_command` a string, but the code below looks like it would work even with a string.

irc_responses.py
```diff
@@ -6,10 +6,14 @@
 ERR_TOOMANYCHANNELS = ("405", ":You have joined too many channels")
 ERR_UNKNOWNCOMMAND = ("421", ":Unknown command")
 ERR_NOMOTD = ("422", ":MOTD File is missing")
+ERR_USERNOTINCHANNEL = ("441", ":They aren't on that channel")
 ERR_NOTONCHANNEL = ("442", ":You're not on that channel")
 ERR_NOTREGISTERED = ("451", ":You have not registered")
+ERR_CHANOPRIVSNEEDED = ("482", ":You're not channel operator")
+ERR_UMODEUNKNOWNFLAG = ("501", ":Unknown MODE flag")
```
Akuli, line 13: It would be useful to see which flag was unknown, e.g. `Unknown MODE flag v`. (Imagine copy/pasting a command that sets 10 different flags and getting this error)

command.py
```diff
@@ -173,6 +184,60 @@ def motd(motd_content: Optional[Dict[str, List[str]]], user: mantatail.UserConne
...
+                        pass
+                    elif mode_command[0] == "+":
+                        state.channels[target_chan.lower()].set_operator(target_user)
+                        message = f"MODE {target_chan} {args[1]} {target_user}"
+                        for receiver in [
+                            user,
+                            state.channels[target_chan.lower()].user_dict[target_user.lower()],
+                        ]:
+                            receiver.send_string_to_client(message)
+                    elif mode_command[0] == "-":
```
Akuli, line 224: What happens if it's neither `+` nor `-`? It should probably be an error.

mantatail.py
```diff
@@ -108,12 +108,22 @@ def send_string_to_client(self, message: str, prefix: str = "mantatail") -> None
 
 
 class Channel:
-    def __init__(self, channel_name: str, channel_creator: str) -> None:
+    def __init__(self, channel_name: str, channel_founder: str) -> None:
```
Akuli, line 111: Why is founder a string, and not a `User` object? A string works, but creates one more gotcha for nick changes.

command.py
```diff
@@ -173,6 +184,60 @@ def motd(motd_content: Optional[Dict[str, List[str]]], user: mantatail.UserConne
...
+
+        command_contains_unsupported_modes = False
+
+        for mode in mode_command[1:]:
+            if mode not in supported_channel_modes:
+                command_contains_unsupported_modes = True
+            else:
+                if mode == "o":
+                    if user.nick != state.channels[target_chan.lower()].founder:
+                        error_no_operator_privileges(user, target_chan)
```
Akuli, line 212: Should operators be allowed to add more operators?

command.py
```diff
@@ -173,6 +184,60 @@ def motd(motd_content: Optional[Dict[str, List[str]]], user: mantatail.UserConne
...
+                    elif target_user.lower() not in state.channels[target_chan].user_dict.keys():
+                        error_user_not_in_channel(user, target_user, target_chan)
+                        pass
```
Akuli, line 215:
```suggestion
```

command.py
```diff
@@ -173,6 +184,60 @@ def motd(motd_content: Optional[Dict[str, List[str]]], user: mantatail.UserConne
...
+        mode_command = list(args[1])
+        target_user = args[2]
+
+        command_contains_unsupported_modes = False
+
+        for mode in mode_command[1:]:
+            if mode not in supported_channel_modes:
+                command_contains_unsupported_modes = True
+            else:
+                if mode == "o":
```
Akuli, line 210: I don't really get the purpose of `supported_channel_modes`. You could just do this:

```python3
if mode == "o":
    ...
else:
    command_contains_unsupported_modes = True
```

This way, the if statements contain all the info you need, and you don't need to duplicate it to the `supported_channel_modes` list.

command.py
```diff
@@ -173,6 +184,60 @@ def motd(motd_content: Optional[Dict[str, List[str]]], user: mantatail.UserConne
...
+                    elif target_user.lower() not in state.channels[target_chan].user_dict.keys():
+                        error_user_not_in_channel(user, target_user, target_chan)
+                        pass
+                    elif mode_command[0] == "+":
+                        state.channels[target_chan.lower()].set_operator(target_user)
+                        message = f"MODE {target_chan} {args[1]} {target_user}"
+                        for receiver in [
+                            user,
+                            state.channels[target_chan.lower()].user_dict[target_user.lower()],
+                        ]:
```
Akuli, line 222: Shouldn't everyone on the channel notice when there is a new operator? (Alice and Bob might get a new friend Charley.)

mantatail.py
```diff
@@ -108,12 +108,22 @@ def send_string_to_client(self, message: str, prefix: str = "mantatail") -> None
...
+        self.operators: Set[str] = set()
         self.user_dict: Dict[Optional[str], UserConnection] = {}
 
+        self.operators.add(self.founder.lower())
+
+    def set_operator(self, user: str) -> None:
+        self.operators.add(user)
+
+    def remove_operator(self, user: str) -> None:
+        self.operators.discard(user)
```
Akuli, line 125: This should run when an operator leaves (PART or QUIT) the channel.

Akuli (commented):
One more thought:

mantatail.py
Akuli: I think this is going to be bad either way. If it's the nick as a string and the founder QUITs, anyone can "become" the founder with just a nick change. If it's a User object, that user will not exist anywhere else once the founder QUITs, and nobody can have channel founder permissions anymore.

command.py
ThePhilgrim: No, I don't think so. At least I designed this with the intention of only letting the founder set operator status

command.py
```diff
@@ -7,31 +7,24 @@
...
             if lower_user_nick not in state.channels[lower_channel_name].user_dict.keys():
 
                 channel_user_keys = state.channels[lower_channel_name].user_dict.keys()
                 channel_users = " ".join(
-                    [
-                        state.channels[lower_channel_name].user_dict[user_key].nick
-                        for user_key in channel_user_keys
-                    ]
+                    [state.channels[lower_channel_name].user_dict[user_key].nick for user_key in channel_user_keys]
                 )
```
Akuli, line 28: `.values()` would make this a lot easier.

command.py
```diff
@@ -80,6 +71,9 @@ def handle_part(
         else:
             channel_users = state.channels[lower_channel_name].user_dict
 
+            if lower_user_nick in state.channels[lower_channel_name].operators:
+                state.channels[lower_channel_name].remove_operator(user.nick.lower())
```
Akuli, line 75: There should probably be a test for this.

command.py
```diff
@@ -173,6 +179,50 @@ def motd(motd_content: Optional[Dict[str, List[str]]], user: mantatail.UserConne
...
+                if mode == "o":
+                    if user.nick.lower() not in state.channels[target_chan.lower()].operators:
+                        error_no_operator_privileges(user, target_chan)
+                    elif target_user.lower() not in state.channels[target_chan].user_dict.keys():
+                        error_user_not_in_channel(user, target_user, target_chan)
+                    elif mode_command[0] == "+":
+                        state.channels[target_chan.lower()].set_operator(target_user)
+                        message = f"MODE {target_chan} {args[1]} {target_user}"
+                        for receiver_nick, receiver in state.channels[target_chan.lower()].user_dict.items():
+                            receiver.send_string_to_client(message)
```
Akuli, line 208: Nit: `.values()` is enough, because `receiver_nick` is not used.

tests/test_mantatail.py
```diff
@@ -205,6 +208,52 @@ def test_send_unknown_commands(user_alice):
...
+        pass
+
+    user_bob.sendall(b"MODE #foo +o Alice\r\n")
+    received = receive_line(user_bob)
+    assert received == b":mantatail 482 #foo :You're not channel operator\r\n"
+
+
+def test_operator_user_not_in_channel(user_alice, user_bob):
+    user_alice.sendall(b"JOIN #foo\r\n")
+    user_bob.sendall(b"JOIN #bar\r\n")
```
Akuli, line 246: Bob doesn't have to join any channel for this test to work:
```suggestion
```

command.py
ThePhilgrim: Awesome! I didn't know there was a `.values()` for dicts! Much needed

ThePhilgrim:
@Akuli Updated! I'm on IRC if you want to come :)

command.py
```diff
@@ -7,43 +7,43 @@
...
-                channel_users = " ".join(
-                    [
-                        state.channels[lower_channel_name].user_dict[user_key].nick
-                        for user_key in channel_user_keys
-                    ]
-                )
+
+                channel_user_nicks = []
+                for channel_user in state.channels[lower_channel_name].user_dict.values():
+                    if channel_user.nick.lower() in state.channels[lower_channel_name].operators:
```
Akuli, line 29: `state.channels[lower_channel_name]` is repeated a lot. Maybe set it to a variable?

command.py
```diff
@@ -7,43 +7,43 @@
...
         else:
             if lower_channel_name not in state.channels.keys():
-                state.channels[lower_channel_name] = mantatail.Channel(channel_name, user.nick)
+                state.channels[lower_channel_name] = mantatail.Channel(channel_name, user)
 
             lower_user_nick = user.nick.lower()
 
             if lower_user_nick not in state.channels[lower_channel_name].user_dict.keys():
 
                 channel_user_keys = state.channels[lower_channel_name].user_dict.keys()
```
Akuli, line 25: What's the purpose of this variable? It's used only once, and `.values()` would be enough for that.

tests/test_mantatail.py
```diff
@@ -205,6 +227,102 @@ def test_send_unknown_commands(user_alice):
...
+    while receive_line(user_alice) != b":Bob!BobUsr@127.0.0.1 JOIN #foo\r\n":
+        pass
+    while receive_line(user_bob) != b":mantatail 366 Bob #foo :End of /NAMES list.\r\n":
+        pass
+
+    user_alice.sendall(b"MODE #foo +o Bob\r\n")
+    received = receive_line(user_alice)
+    assert received == b":mantatail MODE #foo +o Bob\r\n"
+    received = receive_line(user_bob)
+    assert received == b":mantatail MODE #foo +o Bob\r\n"
```
Akuli, line 255: Nit: You can write asserts like this as e.g. `assert receive_line(user_alice) == b"blah blah\r\n"`. This can make the tests easier to read, especially when there is many asserts.

PR #38, opened 2022-01-01
Implement logic to prevent duplicate nicks

### LINKED ISSUE #36: duplicate nicks
2022-01-01-125829_1280x800_scrot

Users are often identified by nicknames. This isn't great, if two people with nickname `x` connect at the same time. The second person connecting should get an error saying that the nickname is already in use.

ThePhilgrim:
Fixes #36

### DIFF
--- a/command.py
+++ b/command.py
@@ -113,6 +113,8 @@ def handle_quit(state: mantatail.ServerState, user: mantatail.UserConnection, co
         for receiver in receivers:
             receiver.send_string_to_client(message, prefix=user.user_mask)
 
+        delete_user_from_server_users(state, user)
+
         user.closed_connection = True
         user.socket.close()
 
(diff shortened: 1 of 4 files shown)


command.py
```diff
@@ -173,6 +175,12 @@ def motd(motd_content: Optional[Dict[str, List[str]]], user: mantatail.UserConne
     user.send_string_to_client(motd_start_and_end["end_msg"])
 
 
+def delete_user_from_server_users(
+    state: mantatail.ServerState, user: mantatail.UserConnection
+) -> None:
+    state.users.discard(user.nick.lower())
```
Akuli, line 181: - This doesn't delete the user from each joined channel. This is fine when it's called from `handle_quit()`, which does that before the call, but this is not fine when it's called directly from `mantatail.py`.
- Should this be a method of `state`?

mantatail.py
```diff
@@ -11,6 +11,7 @@ class ServerState:
     def __init__(self, motd_content: Optional[Dict[str, List[str]]]) -> None:
         self.lock = threading.Lock()
         self.channels: Dict[str, Channel] = {}
+        self.users: Set[str] = set()
```
Akuli, line 14: Why is this a list of strings, and not `User` objects?

tests/test_mantatail.py
```diff
@@ -215,6 +215,35 @@ def test_connect_via_netcat(run_server):
...
 
+def test_nick_already_taken(run_server):
+    with socket.socket() as nc:
+        nc.connect(("localhost", 6667))
+        nc.sendall(b"NICK nc\n")
+        nc.sendall(b"USER nc 0 * :netcat\n")
+        while receive_line(nc) != b":mantatail 376 nc :End of /MOTD command\r\n":
+            pass
+
+    with socket.socket() as nc2:
```
Akuli, line 226: The socket `nc` closes when the `with` statement ends, so the first `nc` user might not exist here anymore. This is also why the `QUIT` messages aren't working.

PR #39, opened 2022-01-02
Restructure how server store users

### LINKED ISSUE #36: duplicate nicks

ThePhilgrim:
Fixes #36


command.py
```diff
@@ -50,44 +46,30 @@ def handle_join(state: mantatail.ServerState, user: mantatail.UserConnection, ch
...
-    # TODO: Show part message to other users & Remove from user from channel user list.
-    lower_channel_name = channel_name.lower()
-    lower_user_nick = user.nick.lower()
+    try:
+        channel = state.channels[channel_name.lower()]
+    except KeyError:
+        error_no_such_channel(user, channel_name)
+        return
 
     with state.lock:
```
Akuli, line 59: Should probably hold the lock while getting the channel, although it's unlikely (impossible?) that the channel goes away while the user hasn't parted it yet.

command.py
```diff
@@ -50,44 +46,30 @@ def handle_join(state: mantatail.ServerState, user: mantatail.UserConnection, ch
...
-                receiver.send_string_to_client(message, prefix=user.user_mask)
+                usr.send_string_to_client(message, prefix=user.user_mask)
 
-            del channel_users[lower_user_nick]
-
-            if len(state.channels[lower_channel_name].user_dict) == 0:
-                del state.channels[lower_channel_name]
+            channel.users.discard(user)
+            if len(channel.users) == 0:
+                del channel
```
Akuli, line 72: `del channel` deletes a local variable. Nothing else. Maybe you meant `del state.channels[channel_name.lower()]`?

One possible way to write a test that ensures this is correct:
- Alice joins `#foo`
- Alice parts `#foo`
- Bob joins `#foo`
- Assert that the founder of `#foo` is Bob (using some command that returns information about the channel)

If the channel isn't deleted when Alice parts, the founder of the channel will still be Alice. Instead of the founder, you can use any other property of channels that should not be preserved when the channel is deleted and recreated.

command.py
```diff
@@ -185,6 +168,12 @@ def motd(motd_content: Optional[Dict[str, List[str]]], user: mantatail.UserConne
 
 
 def process_channel_modes(state: mantatail.ServerState, user: mantatail.UserConnection, args: List[str]) -> None:
+    supported_modes = ["+", "-", "o"]
```
Akuli, line 171: `o+` is not a valid mode string, and should be an error instead of doing something weird.

tests/test_mantatail.py
```diff
@@ -141,6 +141,11 @@ def receive_line(sock):
     return received
 
 
+# Makes it easier to assert bytes received from Sets
+def compare_sets(received_bytes, compare_with):
```
Akuli, line 145: The function name is vague. It says that the function is implemented by comparing sets, but how is that supposed to help me remember what the function does? How about `same_words()`, `words_match_in_any_order()`, or some other name that describes how it checks whether the two byte strings have the same words?

If the function is named correctly, the comment will probably become redundant.

tests/test_mantatail.py
```diff
@@ -325,3 +342,31 @@ def test_join_part_race_condition(user_alice, user_bob):
...
+        pass
+    nc.close()
+
+    time.sleep(0.1)
+
+    nc2.sendall(b"NICK nc\n")
+    nc2.sendall(b"USER nc\n")
+
+    while receive_line(nc2) != b":mantatail 376 nc :End of /MOTD command\r\n":
+        pass
```
Akuli, line 372: `nc2` is not closed.

Akuli (commented):
I like this overall! The code is more readable than ever before.

command.py
ThePhilgrim: I think it isn't possible. But I guess it doesn't cost a lot to be on the safe side here.

command.py
ThePhilgrim: Ah, when using `del` you're not referring to the thing the variable is referring to? Interesting

ThePhilgrim: > * Assert that the founder of `#foo` is Bob (using some command that returns information about the channel)

At the moment, I have no commands that would give me the channel info.

ThePhilgrim:
Channel owner prefix based on [https://modern.ircdocs.horse/#founder-prefix](Irc Horse Founder Prefix)

command.py
```diff
@@ -185,6 +170,15 @@ def motd(motd_content: Optional[Dict[str, List[str]]], user: mantatail.UserConne
 
 
 def process_channel_modes(state: mantatail.ServerState, user: mantatail.UserConnection, args: List[str]) -> None:
+    supported_modes = ["o"]
+    for mode in args[1][1:]:
+        if args[1][0] not in ["+", "-"]:
+            error_unknown_mode(user, args[1][0])
+            return
```
Akuli, line 177: Having the `args[1][0]` check inside a loop feels weird. It's not wrong, but it would make more sense outside the loop.

PR #40, opened 2022-01-03
Kick (Based on branch user_logic)

ThePhilgrim:
TODO:
- [x] Write tests
- [x] Implement functionality for added Reason (`KICK #chan user :Reason`)
- [x] Check if operators can kick operators/founder


Akuli:
Operators kicking operators: Yes. I can even kick myself from ##learnpython.

command.py
```diff
@@ -83,9 +83,33 @@ def handle_mode(state: mantatail.ServerState, user: mantatail.UserConnection, mo
         process_user_modes()
 
 
-# !Not implemented
-def _handle_kick(message: str) -> None:
-    pass
+def handle_kick(state: mantatail.ServerState, user: mantatail.UserConnection, arg: str) -> None:
+    args = arg.split()
```
Akuli, line 87: If the kick reason contains repeated spaces, this will treat each repeated space as if it was just a single space. Not a big deal if
```
you are  a    troll
```
turns into
```
you are a troll
```
but we might as well fix this.

```suggestion
    args = arg.split(" ")
```

command.py
```diff
@@ -83,9 +83,33 @@ def handle_mode(state: mantatail.ServerState, user: mantatail.UserConnection, mo
...
+    ):
+        error_user_not_in_channel(user, state.connected_users[args[1].lower()], state.channels[args[0].lower()])
+    else:
+        if len(args) == 2:
+            message = f"KICK {state.channels[args[0].lower()].name} {state.connected_users[args[1].lower()].nick}\r\n"
+        elif len(args) >= 3:
+            if not args[2].startswith(":"):
+                reason = f':{" ".join(args[2:])}'
+            else:
+                reason = " ".join(args[2:])
```
Akuli, line 110: `KICK #foo Akuli x y` should be an error, if I don't put `:` in front of `x`.

command.py
```diff
@@ -294,7 +318,7 @@ def error_unknown_mode(user: mantatail.UserConnection, unknown_command: str) ->
     user.send_string_to_client(message)
 
 
-def error_not_enough_params(user: mantatail.UserConnection, target_chan: str) -> None:
+def error_not_enough_params(user: mantatail.UserConnection, target_chan_nick: str, command: str) -> None:
```
Akuli, line 321: Should the argument be named `target_chan_name`? Nicks are a user thing, not a channel thing.

mantatail.py
```diff
@@ -130,6 +130,13 @@ def set_operator(self, user_nick_lower: str) -> None:
     def remove_operator(self, user_nick_lower: str) -> None:
         self.operators.discard(user_nick_lower)
 
+    def kick_user(self, kicker: UserConnection, user_to_kick: UserConnection, message: str) -> None:
+        # TODO: Can an operator kick founder/another operator?
```
Akuli, line 134: Yes, they can, and the code handles it already. Kicking yourself also should work with this code, but it would be good to add a test ensuring it works. Then the TODO comment can be deleted.

tests/test_mantatail.py
```diff
@@ -358,6 +358,50 @@ def test_operator_user_not_in_channel(user_alice, user_bob):
...
+        pass
+
+    user_bob.sendall(b"JOIN #foo\r\n")
+
+    while receive_line(user_alice) != b":Bob!BobUsr@127.0.0.1 JOIN #foo\r\n":
+        pass
+    while receive_line(user_bob) != b":mantatail 366 Bob #foo :End of /NAMES list.\r\n":
+        pass
+
+    user_alice.sendall(b"KICK #foo Bob Bye bye\r\n")
```
Akuli, line 387: This should be an error. `Bye bye` is two arguments, but I'm pretty sure the IRC docs say that there should only be one argument? If prefixed with `:`, as in `:Bye bye`, then `Bye bye` is one argument.

tests/test_mantatail.py
```diff
@@ -358,6 +358,50 @@ def test_operator_user_not_in_channel(user_alice, user_bob):
...
+    assert receive_line(user_bob) == b":Alice!AliceUsr@127.0.0.1 KICK #foo Bob :Bye bye\r\n"
+
+    user_bob.sendall(b"JOIN #foo\r\n")
+
+    while receive_line(user_alice) != b":Bob!BobUsr@127.0.0.1 JOIN #foo\r\n":
+        pass
+    while receive_line(user_bob) != b":mantatail 366 Bob #foo :End of /NAMES list.\r\n":
+        pass
+
+    user_alice.sendall(b"KICK #foo Bob :Reason\r\n")
```
Akuli, line 399: The reason is optional. There doesn't seem to be any tests for kicking without specifying a reason.

PR #42, opened 2022-01-03
Find / Delete object methods + improve readability

ThePhilgrim:


command.py
```diff
@@ -88,28 +88,36 @@ def handle_kick(state: mantatail.ServerState, user: mantatail.UserConnection, ar
...
+        target_usr = state.find_user(args[1])
+    except KeyError:
+        error_no_such_nick_channel(user, args[1])
+        return
+
+    if user.nick.lower() not in channel.operators:
+        error_no_operator_privileges(user, state.find_channel(args[0]))
+        return
+
+    if len(args) >= 2 and target_usr in state.connected_users.values() and target_usr not in channel.users:
```
Akuli, line 108: Because target_usr was returned from `find_user()`, it is guaranteed to be in `state.connected_users`. Also, there will be at least 2 arguments, because `error_not_enough_params` was handled above.

```suggestion
    if target_usr not in channel.users:
```

command.py
```diff
@@ -203,19 +211,22 @@ def process_channel_modes(state: mantatail.ServerState, user: mantatail.UserConn
             error_unknown_mode(user, mode)
             return
 
+    try:
+        channel = state.find_channel(args[0])
+    except KeyError:
+        error_no_such_channel(user, args[0])
+        return
+
     with state.lock:
```
Akuli, line 220: Finding the channel should be done while holding the lock. You can `return` inside a `with` statement, and it will do the right thing.

mantatail.py
```diff
@@ -14,6 +14,18 @@ def __init__(self, motd_content: Optional[Dict[str, List[str]]]) -> None:
...
         self.motd_content = motd_content
 
+    def find_user(self, nick: str) -> UserConnection:
+        return self.connected_users[nick.lower()]
+
+    def find_channel(self, channel_name: str) -> Channel:
+        return self.channels[channel_name.lower()]
+
+    def delete_user(self, nick: str) -> None:
+        del self.connected_users[nick.lower()]
```
Akuli, line 24: The user should also be removed from any channels they have joined. You don't want in any case a channel to have a user that isn't in `state.connected_users`. But the PR is already big enough as is, I would just add a todo comment about this.

Akuli:
`command.py` still contains several `lower()` calls:

```
akuli@akuli-desktop:~/MantaTail$ grep "lower()" command.py
    lower_channel_name = channel_name.lower()
                    elif usr.nick.lower() in channel.operators:
            channel = state.channels[channel_name.lower()]
            if user.nick.lower() in channel.operators:
                channel.remove_operator(user.nick.lower())
    if user.nick.lower() not in channel.operators:
            if user.nick.lower() in channel.operators:
                channel.remove_operator(user.nick.lower())
                    if user.nick.lower() not in channel.operators:
                        channel.set_operator(target_usr.nick.lower())
                        channel.remove_operator(target_usr.nick.lower())
```

It would be nice if more of these were replaced with methods that do the `.lower()` automatically. When you want to write code for individual IRC commands, you shouldn't have to worry about how exactly nicks are stored in data structures; that's a responsibility of the `ServerState` class.

command.py
ThePhilgrim: @Akuli I guess this will be in the lock when the other published PR is merged. It has the lock around call_handler_function(). All work here is done in that lock I believe

mantatail.py
ThePhilgrim: This is just a quick fix. Will implement it in coming push.

ThePhilgrim:
I have not made methods that take care of most `.lower()`. The ones remaining I think it makes sense to leave them. They are strings. I wouldn't necessarily think it's reasonable to have `channel_name`, convert it to the channel, and then get the lower channel name. In these cases it makes sense to just go `.lower() on them`

PR #46, opened 2022-01-04
Working on implementing a send que. Code seems to work, tests error

ThePhilgrim:


mantatail.py
```diff
@@ -62,15 +63,14 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
...
+                    user.send_que.put((None, None))
+                    return
                 if request_chunk:
                     request += request_chunk
                 else:
-                    if user is not None:
-                        print(f"{user.nick} has disconnected.")
-                        state.delete_user(user.nick)
-                    else:
-                        print("Disconnected.")
```
Akuli, line 73: There are many ways how a user might disconnect, and this is one of them, as the deleted prints suggest. This one happens when you press Ctrl+D in netcat. So this too should put `(None, None)` to the queue.

To write a test for this, you can call `socket.close()` somewhere during the test randomly, and make sure that a different user gets a QUIT message.

mantatail.py
```diff
@@ -110,25 +110,55 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
...
         self.socket.sendall(message_as_bytes)
 
+    def send_quit_message(self) -> None:
+        # TODO: Implement logic for different reasons & disconnects.
+        reason = "(Remote host closed the connection)"
+        message = f"QUIT :Quit: {reason}"
+
+        receivers = set()
+        receivers.add(self)
+        for channel in self.state.channels.values():
```
Akuli, line 151: Should probably hold the lock.

mantatail.py
```diff
@@ -110,25 +110,55 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
...
+            if self in channel.users:
+                for usr in channel.users:
+                    receivers.add(usr)
+
+            if channel.is_operator(self):
+                channel.remove_operator(self)
+
+        for receiver in receivers:
+            receiver.send_string_to_client(message, self.user_mask)
```
Akuli, line 160: - For other users than self, this should add to send queue instead of directly sending. Think about what happens if one user has slow internet.
- Sending the quit message to self is trickier. We can't add to the send queue, because when the sending thread calls this method, it is about to return and never `.get()` again. We can call `send_string_to_client()`, but we should do it after everything else in case it's slow, and also be prepared to handle errors. We can ignore sending errors silently, because the user is QUITting anyway.

mantatail.py
```diff
@@ -110,25 +110,55 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
...
+    def start_queue_listener(self) -> None:
+        while True:
+            (message, prefix) = self.send_que.get()
 
-    def send_string_to_client(self, message: str, prefix: str = "mantatail") -> None:
+            if message is None or prefix is None:
+                self.send_quit_message()
+                self.state.delete_user(self.nick)
+                self.socket.close()
+                return print(f"{self.nick} has disconnected.")
```
Akuli, line 136: Because print always returns `None`, `return print("blah")` is correct, but it's confusing. Just write it on separate lines:
```suggestion
                print(f"{self.nick} has disconnected.")
                return
```

mantatail.py
```diff
@@ -62,15 +63,15 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
             request = b""
             # IRC messages always end with b"\r\n" (netcat uses "\n")
             while not request.endswith(b"\n"):
-                request_chunk = user_socket.recv(4096)
+                try:
+                    request_chunk = user_socket.recv(4096)
+                except OSError:
+                    user.send_que.put((None, None))  # type: ignore
```
Akuli, line 69: This will try to do `None.send_que`, if the user hasn't sent NICK and USER yet when they disconnect. Same goes for the `# type: ignore` below.

mantatail.py
```diff
@@ -110,25 +111,62 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
...
+                if channel.is_operator(self):
+                    channel.remove_operator(self)
+
+            for receiver in receivers:
+                receiver.send_que.put((message, self.user_mask))
+
+            try:
+                self.send_string_to_client(message, self.user_mask)
+            except:
+                print("HELLO NO")
```
Akuli, line 168:
```suggestion
            except OSError:
                pass  # user QUITting already
```

PR #49, opened 2022-01-06
Add thread lock for sending quit msg and deleting user

### LINKED ISSUE #48: Potential race condition when calling delete_user
https://github.com/ThePhilgrim/MantaTail/blob/3b8f22dc089d09d2f5f6f6b8bd64ce627f5bc589/mantatail.py#L130-L136

`delete_user` loops over each channel. These loops will fail if channels are added/deleting during the loop, so the lock should be held while calling `delete_user`.

I also think the lock shouldn't be released between calling `send_quit_message` and calling `delete_user`. Instead, both calls should be done in a single `with lock:`. If the lock is acquired separately for the two calls, and released in between, it is possible for someone to see a user in the state when joining a new channel even though a QUIT message has been sent.

ThePhilgrim:
Fixes #48

Added `self.socket.close()`. I don't see that it would be problematic even though it might not necessarily have to be there.

### DIFF
--- a/mantatail.py
+++ b/mantatail.py
@@ -158,24 +158,23 @@ class UserConnection:
         reason = "(Remote host closed the connection)"
         message = f"QUIT :Quit: {reason}"
 
-        with self.state.lock:
-            receivers = set()
-            # receivers.add(self)
-            for channel in self.state.channels.values():
-                if self in channel.users:
-                    for usr in channel.users:
-                        receivers.add(usr)
+        receivers = set()
+        # receivers.add(self)
+        for channel in self.state.channels.values():
+            if self in channel.users:
+                for usr in channel.users:
+                    receivers.add(usr)
 
-                if channel.is_operator(self):
-                    channel.remove_operator(self)
+            if channel.is_operator(self):
+                channel.remove_operator(self)
 
-            for receiver in receivers:
-                receiver.send_que.put((message, self.user_mask))
+        for receiver in receivers:
+            receiver.send_que.put((message, self.user_mask))
 
-            try:
-                self.send_string_to_client(message, self.user_mask)
-            except OSError:
-                return
+        try:
+            self.send_string_to_client(message, self.user_mask)
+        except OSError:
+            return
 
 
 class Channel:
--- a/mantatail.py
+++ b/mantatail.py
@@ -124,18 +124,26 @@ class UserConnection:
         self.user_message = user_message
         self.user_name = user_message.split(" ", 1)[0]
         self.user_mask = f"{self.nick}!{self.user_name}@{self.host}"
-        self.que_thread = threading.Thread(target=self.start_queue_listener)
+        self.que_thread = threading.Thread(target=self.send_queue_thread)
         self.que_thread.start()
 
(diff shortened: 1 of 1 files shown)


mantatail.py
```diff
@@ -132,9 +132,10 @@ def start_queue_listener(self) -> None:
             (message, prefix) = self.send_que.get()
 
             if message is None or prefix is None:
-                self.send_quit_message()
-                self.state.delete_user(self.nick)
-                self.socket.close()
+                with self.state.lock:
+                    self.send_quit_message()
```
Akuli, line 136: Tests are getting stuck because `send_quit_message()` has its own locking, and it can't acquire the lock while it's already being held here. The locking in `send_quit_message()` should be just deleted, and it's no longer needed with locking here.

mantatail.py
```diff
@@ -157,24 +158,23 @@ def send_quit_message(self) -> None:
...
-                if self in channel.users:
-                    for usr in channel.users:
-                        if usr != self:
-                            receivers.add(usr)
+        receivers = set()
+        # receivers.add(self)
+        for channel in self.state.channels.values():
+            if self in channel.users:
+                for usr in channel.users:
+                    receivers.add(usr)
```
Akuli, line 166: Any reason to remove `if usr != self`? The comment seems unnecessary too.

mantatail.py
```diff
@@ -157,24 +158,23 @@ def send_quit_message(self) -> None:
...
-                receiver.send_que.put((message, self.user_mask))
+        for receiver in receivers:
+            receiver.send_que.put((message, self.user_mask))
 
-            try:
-                self.send_string_to_client(message, self.user_mask)
-            except OSError:
-                return
+        try:
+            self.send_string_to_client(message, self.user_mask)
```
Akuli, line 175: This shouldn't be done while holding the lock. Sending can be slow, and while we hold the lock, other threads can't handle commands from other clients.

It might make sense to split `send_quit_message()` into two parts: `queue_sending_quit_to_others()` (run this while holding the lock) and `send_quit_to_self()` (run this after releasing the lock).

I can't think of a good way to test this. It would require `.send()` to be deliberately slowed down, but `.send()` to localhost tends to be immediate, and the data just waits somewhere until it's `.recv()`ed.

mantatail.py
ThePhilgrim: Wtf...

mantatail.py
```diff
@@ -124,16 +124,25 @@ def __init__(self, state: ServerState, host: str, socket: socket.socket, user_me
...
-                self.send_quit_message()
-                self.state.delete_user(self.nick)
+                with self.state.lock:
+                    self.queue_quit_message_for_other_users()
+                    self.state.delete_user(self.nick)
+
+                try:
+                    reason = "(Remote host closed the connection)"
+                    quit_message = f"QUIT :Quit: {reason}"
+                    self.send_string_to_client(quit_message, self.user_mask)
```
Akuli, line 142:
```suggestion
                    # Can be slow, if user has bad internet. Don't do this while holding the lock.
                    self.send_string_to_client(quit_message, self.user_mask)
```

PR #52, opened 2022-01-07
PING message

ThePhilgrim:
This is a draft of a PR to start a discussion on my thoughts on how to implement a `PING` message sent from server to client.

As it is now, the server will send a `PING` message to the client every 10 seconds unless there is any communication from the client. Any communication from the client will reset this timer.
10 seconds is of course an unreasonable interval, this is just for developing the functionality.

Still to implement:
- [ ] Proper placement of `.put()` and manage to remove `if user:` checks.
- [ ] Assert that client returns a `PONG :mantatail` within a given time after server sends a `PING`.

@Akuli: Looking for your thoughts on this way of implementing this functionality, and if it's good, what do I need to consider?

### DIFF
--- a/mantatail.py
+++ b/mantatail.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 import socket
 import threading
+import time
 import queue
 import json
 from typing import Dict, Optional, List, Set, Tuple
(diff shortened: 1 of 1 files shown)


Akuli:
`threading.Timer` should make this a lot easier.

ThePhilgrim:
YAY! That's brilliant! I didn't know it existed, it even has a `.cancel()`!

Cred to be for the attempt of a brute-force solution, though? xD

PR #53, opened 2022-01-09
Implement logic to send PING to clients

ThePhilgrim:
A ping message is currently sent to a user who has been inactive for 10 minutes, (`PING :mantatail`). 

If the client responds with `PONG :mantatail` within 5 seconds, the timer is restarted. Otherwise the client is disconnected from the server.

Disconnect reasons are still not supported, and the `QUIT` message therefore doesn't distinguish an intentional quit from a disconnect. This feature will be implemented in a future PR.

### DIFF
--- a/mantatail.py
+++ b/mantatail.py
@@ -136,6 +136,7 @@ class UserConnection:
         self.user_mask = f"{self.nick}!{self.user_name}@{self.host}"
         self.que_thread = threading.Thread(target=self.send_queue_thread)
         self.que_thread.start()
+        self.pong_received = False
 
     def send_queue_thread(self) -> None:
         while True:
@@ -195,10 +196,13 @@ class UserConnection:
 
     def queue_ping_message(self) -> None:
         self.send_que.put(("PING :mantatail", "mantatail"))
-        self.pong_timer = threading.Timer(5, self.on_no_pong_received).start()
+        threading.Timer(5, self.assert_pong_received).start()
 
-    def on_no_pong_received(self) -> None:
-        self.send_que.put((None, None))
+    def assert_pong_received(self) -> None:
+        if not self.pong_received:
+            self.send_que.put((None, None))
+        else:
+            self.pong_received = False
 
 
 class Channel:
--- a/mantatail.py
+++ b/mantatail.py
@@ -64,6 +64,10 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
             # IRC messages always end with b"\r\n" (netcat uses "\n")
             while not request.endswith(b"\n"):
                 try:
+                    # Temporary check until these actions can be done before user registration
+                    if user:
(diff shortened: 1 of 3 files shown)


mantatail.py
```diff
@@ -64,6 +64,10 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
             # IRC messages always end with b"\r\n" (netcat uses "\n")
             while not request.endswith(b"\n"):
                 try:
+                    # Temporary check until these actions can be done before user registration
+                    if user:
+                        ping_timer = threading.Timer(600, user.queue_ping_message)
```
Akuli, line 69: I don't understand why this code is placed here. If it happens to take 5 `.recv()` calls to receive a single line of text, why do we start 5 different timers? Why do we then stop all of them? Why don't we stop the timer if the client has disconnected? I'm just really confused. Maybe there's a reason to do exactly what this does; if there is, please explain.

Akuli:
It would be nice to see a test for this. In a test, you obviously don't want to wait 10 minutes, but if you put both wait times (600 and 5) to global variables, you can temporarily change them in a test. See pytest's `monkeypatch` fixture.

mantatail.py
```diff
@@ -171,13 +184,22 @@ def queue_quit_message_for_other_users(self) -> None:
...
 
+    def queue_ping_message(self) -> None:
+        self.send_que.put(("PING :mantatail", "mantatail"))
+        threading.Timer(5, self.assert_pong_received).start()
+
+    def assert_pong_received(self) -> None:
+        if not self.pong_received:
+            self.send_que.put((None, None))
+        else:
+            self.pong_received = False
```
Akuli, line 195: This could be slightly cleaner with `.cancel()`ing a timer: change the timer to do `self.send_que.put((None, None))`, and `.cancel()` it in `handle_pong()`. (What you have works too, and it is actually how timers are implemented.)

mantatail.py
```diff
@@ -64,10 +67,18 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
             # IRC messages always end with b"\r\n" (netcat uses "\n")
             while not request.endswith(b"\n"):
                 try:
+                    # Temporary check until these actions can be done before user registration
+                    if user:
+                        ping_timer = threading.Timer(TIMER_SECONDS, user.queue_ping_message)
+                        ping_timer.start()
```
Akuli, line 73: Starting the timer doesn't need to be inside `try:`. It is not going to fail with `OSError`.

Akuli (commented):
The last nits...

ThePhilgrim:
@Akuli 
Using `user` to cancel the 5 second `pong_timer` introduced some issues (for example, client sends `PONG :mantatail` even if they didn't receive a `PING`, which leads to an `AttributeError`.) There were also issues with the tests timing out with that logic. 

I feel that in the end, the boolean logic works well, is simple and clear.

PR #54, opened 2022-01-10
Add try block to privmsg assertion

### LINKED ISSUE #45: Sudden disconnect doesn't remove user
Skärmavbild 2022-01-04 kl  23 18 34

After `PRIVMSG` was sent without a `:`, the client was disconnected. When trying to connect again with the same nick, it was already taken.

ThePhilgrim:
Fixes #45

### DIFF
--- a/command.py
+++ b/command.py
@@ -123,7 +123,10 @@ def handle_quit(state: mantatail.ServerState, user: mantatail.UserConnection, co
 
 def handle_privmsg(state: mantatail.ServerState, user: mantatail.UserConnection, msg: str) -> None:
     (receiver, colon_privmsg) = msg.split(" ", 1)
-    assert colon_privmsg.startswith(":")
+    try:
+        assert colon_privmsg.startswith(":")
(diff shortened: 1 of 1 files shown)


ThePhilgrim:
```
[21:37]	Akuli	this pr to me is a small occurrence of a bigger problem / design question
[21:37]	Akuli	do we want to be prepared to handle anything whatsoever as inputs? should the server respond with 
                                 clear error messages, or is it fine if something causes an exception?
[21:41]	Akuli	a possible cleanup idea: when receiving, handle the ":" thing once and for all, instead of repeating the ":" 
                                 in every handle_foo() function
[21:41]	Akuli	the spec says that an argument starting with ":" means the rest is joined with spaces into a single 
                                 argument, and this applies to all commands
[21:41]	Akuli	each command would then get a list of already splitted arguments, instead of a string that the command 
                                 has to .split()
```

Closing this PR as the feature needs more consideration on a higher level

PR #56, opened 2022-01-11
Add parse args function to make code cleaner, update send_privmsg test

ThePhilgrim:
`TODO` comments are addressed in #57.


ThePhilgrim:
So, I have no idea what happened. To be able to change variable `verb` to `command`, I changed the name of `command.py` to `commands.py`. I have no clue how it didn't get included in the commit.

Pulled the file from `err_handling` which means that now, this PR includes even #57. What a mess.

PR #57, opened 2022-01-11
Err handling

### LINKED ISSUE #55: Sending `PRIVMSG` without channel or message errors
Not enough arguments to unpack – should instead be IRC errors 412 & 411

ThePhilgrim:
Fixes #55


command.py
```diff
@@ -122,8 +135,17 @@ def handle_quit(state: mantatail.ServerState, user: mantatail.UserConnection, co
 
 
 def handle_privmsg(state: mantatail.ServerState, user: mantatail.UserConnection, msg: str) -> None:
-    (receiver, colon_privmsg) = msg.split(" ", 1)
-    assert colon_privmsg.startswith(":")
+    if msg.lower() == "privmsg":
```
Akuli, line 138: These checks don't make any sense to me. Why is the command sometimes included in the message but sometimes not? Why is it coming to `handle_foo()` functions as a string in the first place, instead of being split before calling a `handle_foo()` function?

command.py
```diff
@@ -140,19 +162,21 @@ def handle_privmsg(state: mantatail.ServerState, user: mantatail.UserConnection,
     else:
         for usr in channel.users:
             if usr.nick != user.nick:
-                message = f"PRIVMSG {receiver} {colon_privmsg}"
+                message = f"PRIVMSG {receiver} :{colon_privmsg}"
```
Akuli, line 165: The variable name needs updating: `colon_privmsg` no longer starts with a colon.

command.py
ThePhilgrim: @Akuli

Please see the following code in `recv_loop()`:

```
            for line in split_on_new_line(decoded_message)[:-1]:
                if " " in line:
                    (verb, message) = line.split(" ", 1)
                else:
                    verb = line
                    message = verb

                verb_lower = verb.lower()
                parsed_command = "handle_" + verb_lower
```

If a client only sends 1 word to the server, as in `"PRIVMSG"`, `message` (that is then passed in as an argument when calling `handle_foo`) is the actual verb. So, in this case, it would be called as `handle_privmsg(state, user, "privmsg")`

In some cases, for example, `QUIT`, sending only one word is fine, however, if this is the case in `handle_privmsg`, there are appropriate errors.

If the client sends more than one word to the server, `(verb, message) = line.split(" ", 1)` above separates the command from the message.

I hope this makes more sense now.

Akuli: Why is the code in `recv_loop()` like that? Why doesn't it call `parse_received_args` instead, and take the first arg it returns to be the command?

ThePhilgrim: @Akuli Ah, that is an interesting way to do it. 
I didn't consider it as `recv_loop` is working well as it is. Surely this will make it more concise though.
Will update with a new commit when I have to possibility.

How about #56 ?

Akuli: I saw this PR first when going through a pile of unread emails. It appears to have much of the same code. Feel free to fix this in #56 if you want to keep the pull requests separate :)

mantatail.py
```diff
@@ -86,53 +86,46 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
...
-        # (None, None) disconnects the user
         self.send_que: queue.Queue[Tuple[str, str] | Tuple[None, None]] = queue.Queue()
         self.socket = socket
         self.host = host
-        # Nick is shown in user lists etc, user_name is not
-        self.nick = nick
-        self.user_message = user_message
-        self.user_name = user_message.split(" ", 1)[0]
+        self.nick = nick  # Nick is shown in user lists etc, user_name is not
+        self.user_message = " ".join(user_message)  # Ex. AliceUsr 0 * Alice
```
Akuli, line 127: Is this used somewhere? Does it have to be saved to self?

Also, you have to be careful if you need to use the `user_message` attribute. Its last part is the real name, and it can contain spaces, even though it doesn't start with `:` here.

ThePhilgrim: I'm planning to look over the instantiation of `UserConnection` in the next PR. This will be handled there.

ThePhilgrim:
@Akuli Your comments are resolved. Please have a look over `commands.py` now when the old file is deleted.

commands.py
```diff
@@ -72,21 +83,22 @@ def handle_part(state: mantatail.ServerState, user: mantatail.UserConnection, ch
...
     else:
         process_user_modes()
 
 
-def handle_kick(state: mantatail.ServerState, user: mantatail.UserConnection, arg: str) -> None:
-    args = arg.split(" ")
-
-    if len(args) == 1:
+def handle_kick(state: mantatail.ServerState, user: mantatail.UserConnection, args: List[str]) -> None:
+    if not args:
```
Akuli, line 98: Please add a test case `user_alice.sendall(b"KICK bob\r\n")` to `test_not_enough_params_error`.

Akuli (commented):
One more bug. Once fixed, this is ready to be merged :)

Akuli:
> So, I have no idea what happened. ... I changed the name of command.py to commands.py. ... What a mess. ... Git is so difficult sometimes, lol

Indeed, git can be very confusing with renamed files. It does detect renames, as long as you delete the original file, and they even show up correctly in e.g. `git status` and `git diff`.

But that's about it for what it gets right with file renames. In typeshed, we once did a pull request that renamed almost all files in the repository. We closed or merged almost all other pull requests first, because we knew that Git wouldn't handle it well.

PR #58, opened 2022-01-13
Disconnecting fixes, add back windows-latest test

### LINKED ISSUE #51: Tests are failing on Windows, disabled in github actions
#46 added a send queue, which somehow causes a weird problem on windows: after sending QUIT to the server, a client can no longer receive.

s

Potential next steps:
- Figure out whether it has to do with the server running on windows, or with the client running on windows. Does it even happen with non-localhost connections?
- See how other servers that support windows handle this.
- Try to figure out what to write to google
- Make a much simpler server, e.g. two commands `HELLO` and `QUIT`, add a send queue to that.

Akuli:
Fixes #51


Akuli:
The PR contains a link to a good blog post that explains the problem in detail. I found it by googling "windows socket send last message before close", but the problem isn't windows-specific. In short, the problem was that `.close()` just doesn't do what we assumed it does: it doesn't always wait until the client has received what the server has sent (response to QUIT) before closing the connection.

To fix this, we can't just `.close()` wherever we want and however we want. We have to be quite careful about it. Specifically, we need to do a couple other things before we can `.close()`.

This means the with statement in `recv_loop()` was just plain wrong, because a `with` statement simply ensures that `.close()` gets called. There's no simple way to tell it to do other things before the `.close()`. In fact, a with statement is really just an easier way to write a try/finally statement:

```
# This
with sock:
    blah

# Is same as this
try:
    blah
finally:
    sock.close()
```

I rewrote it as a `try/finally` in this way and then made it do the other things too, or if the user object has been instantiated, ask `send_queue_thread` to close the connection.

ThePhilgrim:
All tests pass after the timeout fix!
Ready to be merged after function renaming :)

Akuli:
Almost half of the function name was `_and_then_close_socket`. I decided to not include closing into the function, and I also shortened the rest.

mantatail.py
```diff
@@ -55,7 +55,7 @@ def run_server_forever(self) -> None:
             client_thread.start()
 
 
-def wait_for_all_data_to_be_sent_and_then_close_socket(sock: socket.socket) -> None:
+def wait_until_data_is_sent(sock: socket.socket) -> None:
```
ThePhilgrim, line 58: I'm not sure, however, if `wait_until_data_is_sent` is a reasonable function name here as. What the function does in essence is to close the connection to the user. "Waiting until data is sent" is just the way it actually fulfills its purpose"

"Waiting until all data is sent" would more reasonably be a part of the docstring telling a developer _how_ the function does what it does.

ThePhilgrim (changes requested):

mantatail.py
ThePhilgrim: @Akuli 

Isn't it closing the socket on the last row here?

```
def close_socket_cleanly(sock: socket.socket) -> None:
    # https://blog.netherlabs.nl/articles/2009/01/18/the-ultimate-so_linger-page-or-why-is-my-tcp-not-reliable
    try:
        sock.shutdown(socket.SHUT_WR)
        sock.settimeout(10)
        sock.recv(1)  # Wait for client to close the connection
    except OSError:
        # Possible causes:
        #   - Client decided to keep its connection open for more than 10sec.
        #   - Client was already disconnected.
        #   - Probably something else too that I didn't think of...
        pass

    sock.close()
```

Also, thanks for updating the function name! `close_socket_cleanly` works great.

ThePhilgrim (approved):

PR #63, opened 2022-01-13
Instantiate user first thing in `recv_loop`

### LINKED ISSUE #47: Can't `QUIT` before giving nick and user to server
Error `You have not registered` when trying to `QUIT` before giving Nick & User

### LINKED ISSUE #61: It is possible to give two people the same nick
Suppose that Alice and Bob connect to MantaTail, and both say `NICK x`. This is not an error:
m

Then Alice and Bob both send `USER x`. Finally, Alice and Bob both send `JOIN #foo`. Now `#foo` now has two different users on it, both named `x`, one as the founder (`~` prefix) and one as not the founder.

m2

I didn't explore what happens if a third user joins in and tries to send a message to nickname `x`. This is already bad enough to be considered a bug :)

ThePhilgrim:
I'm here trying to instantiate `UserConnection` when starting the `recv_loop`. This should hopefully eliminate a bunch of `if user` checks, and generally simplify the code in `recv_loop`.

~~Currently one test failing. Not yet sure why~~

Fixes #47 
Fixes #61


ThePhilgrim:
Fixes #47 

PR currently has mypy issues to be fixed.

ThePhilgrim:
@Akuli My last commit tries to silence some mypy issues. However, there has to be a better way to do this. It's complaining about since I made `user.nick`, `user.user_mask`, & `user.user_name` to `Optional`.

I still get the following mypy output:
```
❯ mypy --strict *.py
commands.py:35: error: Incompatible types in assignment (expression has type "Optional[str]", variable has type "str")
commands.py:42: error: Argument 1 to "put" of "Queue" has incompatible type "Tuple[str, Optional[str]]"; expected "Union[Tuple[str, str], Tuple[None, None]]"
commands.py:79: error: Argument 1 to "put" of "Queue" has incompatible type "Tuple[str, Optional[str]]"; expected "Union[Tuple[str, str], Tuple[None, None]]"
commands.py:162: error: Argument 1 to "put" of "Queue" has incompatible type "Tuple[str, Optional[str]]"; expected "Union[Tuple[str, str], Tuple[None, None]]"
Found 4 errors in 1 file (checked 3 source files)
```

How would you go about fixing this?

Akuli:
`user.nick` probably shouldn't be optional. You need nicks everywhere, and `assert user.nick is not None` is messy. In fact, this is exactly what we wanted to avoid in #13 when we decided to instantiate the user object later. In https://github.com/ThePhilgrim/MantaTail/issues/61#issuecomment-1012536177, I meant that the username can be optional even if the nick isn't.

ThePhilgrim:
@Akuli Okay, so in other words, do you think I should instantiate `UserConnection` in the `if command == "nick` check?
Or is it better to keep it as it was in the beginning?

It's slightly weird, however, to let mypy dictate how the code should be written here. I mean, before nick & user is given, the client should be completely blocked by the error `not registered` which means that there will never be a check where `nick` is `None`.

The code in this PR lets the user `QUIT` properly even if they are not registered. I also get rid of all the redundant `if user` checks.

Akuli:
> do you think I should instantiate UserConnection in the `if command == "nick` check?

This is what I had in mind. But:

> It's slightly weird, however, to let mypy dictate how the code should be written here. I mean, before nick & user is given, the client should be completely blocked by the error not registered which means that there will never be a check where nick is None.

This makes sense, and apart from mypy, just instantiating the user object early (as the PR currently does) simplifies other things, like closing the socket in `finally:`. To silence the mypy warnings once and for all, you can do something like this:

```python3
class UserConnection:
    # Tell mypy that there is a nick attribute, and it's never None
    nick: str

    def __init__(self, unrelated arguments):
        # Set other attributes. DO NOT set self.nick.
        ...

def recv_loop():
    user = UserConnection(...)
    ...
    while True:
        ...
        if blah blah:
            user.nick = something
```

Now mypy thinks that there is a `nick` attribute, because we told it so. But in fact, a newly instantiated user object doesn't have a `nick` attribute. It is set later when the user sends `NICK` for the first time. If you need to check whether it exists, you can use `hasattr(user, "nick")`.

ThePhilgrim:
@Akuli I am back in Italy and wont have the possibility to be on IRC as much as I've been the last month. We can continue communicating here on github. Is there any way to communicate here if I have any questions unrelated to a PR or a project issue?

For example, yesterday I tried to connect to mantatail (running on my computer) from my girlfriend's computer via netcat to see if it can be run via two computers on the same network. Ifconfig on my computer says my IP on the network is `192.168.1.91`. However, when I run `nc 192.168.1.91 6667` from my girlfriend's terminal, nothing happens (with mantatail running ofc).

How would I make it possible to connect via several computers to mantatail running on my computer?

Also, since this question is not related to this PR, is there a place on github we can discuss general things like this?

Akuli:
>is there a place on github we can discuss general things like this?

There is GitHub discussions. You can enable it somewhere in the settings.

Continuing off-topic, a few possible debugging steps for `nc`'ing from another computer:
- Try to ping the computer: `ping 192.168.1.91`. If the computers find each other with that IP, it should print a bunch of stuff; try `ping localhost` or `ping google.com` to see what it should look like. If nothing is printed, or you get error messages, make sure the computers are connected to the same network (e.g. wifi) and the IP is correct. It could also be that if it's hotel wifi or something, it blocks computers from connecting to each other.
- If pinging worked, try passing `-v` (verbose) to netcat, as in `nc -v 192.168.1.91 6667`.

I don't know much about networking details though, so if those steps don't help, I don't know how to troubleshoot this further.

ThePhilgrim:
@Akuli After setting `nick: str` & `user_mask: str` in `class UserConnection`, several "if self.nick" checks fail. For example:

```
❯ python3 -m mantatail
Mantatail running (127.0.0.1:6667)
Exception in thread Thread-1:
Exception in thread Thread-2:
  File "/usr/local/Cellar/python@3.9/3.9.7_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/threading.py", line 973, in _bootstrap_inner
  File "/usr/local/Cellar/python@3.9/3.9.7_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/threading.py", line 973, in _bootstrap_inner
    self.run()
    self.run()
  File "/usr/local/Cellar/python@3.9/3.9.7_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/threading.py", line 910, in run
  File "/usr/local/Cellar/python@3.9/3.9.7_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/threading.py", line 910, in run
  File "/Users/Writing/pthn/mantatail/mantatail.py", line 162, in send_queue_thread
  File "/Users/Writing/pthn/mantatail/mantatail.py", line 101, in recv_loop
    if not user.nick or not user.user_message:
    if self.nick:
```

Code on line 101:
```
if not user.nick or not user.user_message:
                    if command_lower == "user":
                        user.user_message = args
                        user.user_name = args[0]
                        user.generate_user_mask()
```

Code on line 162:
```
if self.nick:
                        self.state.delete_user(self.nick)
```

This was not an issue when the `__init__` method had `self.nick = None`.


Is there a better way to silence mypy regarding this?

mantatail.py
```diff
@@ -104,20 +98,24 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
                 command_lower = command.lower()
                 parsed_command = "handle_" + command_lower
 
-                if user is None:
+                if not user.nick or not user.user_message:
```
Akuli, line 101: Currently all tests fail, because `if not user.nick` fails with `AttributeError` when the `nick` attribute doesn't exist. GitHub Actions didn't run for whatever reason, but I ran `python3 -m pytest` on my computer to check this.

A couple possible fixes, choose your favorite:
- When the nick isn't given yet, use an empty string instead of a non-existent attribute (would simplify other things too). An empty nickname isn't valid, so it's fine to treat it specially.
- Change `if not user.nick` to `if not hasattr(user, "nick")`.

mantatail.py
```diff
@@ -130,26 +128,29 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
...
-        self.user_mask = f"{self.nick}!{self.user_name}@{self.host}"
+        self.user_message: Optional[List[str]] = None  # Ex. AliceUsr 0 * Alice
+        self.user_name: Optional[str] = None  # Ex. AliceUsr
+        self.send_que: queue.Queue[Tuple[str, str] | Tuple[None, None]] = queue.Queue()
         self.que_thread = threading.Thread(target=self.send_queue_thread)
         self.que_thread.start()
         self.pong_received = False
 
+    def generate_user_mask(self) -> None:
+        self.user_mask = f"{self.nick}!{self.user_name}@{self.host}"
```
Akuli, line 152: This method should probably just return the mask, instead of saving it to self. This way the same data isn't in two places.

mantatail.py
```diff
@@ -158,18 +159,21 @@ def send_queue_thread(self) -> None:
             if message is None or prefix is None:
                 with self.state.lock:
                     self.queue_quit_message_for_other_users()
-                    self.state.delete_user(self.nick)
+                    if self.nick:
```
Akuli, line 162: This has the same `AttributeError` problem as the similar code in `recv_loop()` has.

mantatail.py
```diff
@@ -197,7 +201,10 @@ def queue_quit_message_for_other_users(self) -> None:
 
     def send_string_to_client(self, message: str, prefix: str) -> None:
         try:
-            message_as_bytes = bytes(f":{prefix} {message}\r\n", encoding="utf-8")
+            if prefix == "unregistered_user":
```
Akuli, line 204: It would be cleaner to use `None` as the special value.

mantatail.py
ThePhilgrim: Ah, I really like the `hasattr()` solution! Very clean.

mantatail.py
ThePhilgrim: Okay, that would mean that each time I need to send a message where the `user_mask` is the prefix, I need to call this method. Do you think that is a better solution?

Akuli: That's what I would do.

ThePhilgrim:
@Akuli I think this is ready for a final review!

mantatail.py
```diff
@@ -104,20 +98,23 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
...
+                            commands.error_nick_in_use(user, args[0])
                         else:
-                            _nick = args[0]
+                            user.nick = args[0]
                     else:
-                        user_socket.sendall(commands.error_not_registered())
+                        if command_lower == "quit":
+                            user.send_que.put((None, None))
+                        else:
+                            commands.error_not_registered(user)
```
Akuli, line 114: - Can be simplified with `elif`
- Apparently this fixes #47. I'll edit it to the description of the PR, so that #47 is closed when this is merged. (I added it once already, then removed it because I wasn't sure if this really fixes #47.)

mantatail.py
```diff
@@ -130,26 +127,29 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
...
-            user.send_que.put((None, None))
+        user.send_que.put((None, None))
 
 
 class UserConnection:
-    def __init__(self, state: ServerState, host: str, socket: socket.socket, user_message: List[str], nick: str):
+    # self.nick is defined in recv_loop()
+    # It is not set in __init__ to keep mypy happy.
+    nick: str  # Nick is shown in user lists etc, user_name is not
+    user_mask: str  # Ex. Alice!AliceUsr@127.0.0.1
```
Akuli, line 137:
```suggestion
```

mantatail.py
```diff
@@ -72,26 +72,20 @@ def close_socket_cleanly(sock: socket.socket) -> None:
...
 
     try:
         while True:
             request = b""
             # IRC messages always end with b"\r\n" (netcat uses "\n")
             while not request.endswith(b"\n"):
-                # Temporary check until these actions can be done before user registration
-                if user:
-                    user.start_ping_timer()
+                user.start_ping_timer()
```
Akuli, line 82: This can send PING before registering is ready, but cannot receive PONG until the user is registered.

mantatail.py
Akuli: Can you also add a test for QUITting before you are registered?

ThePhilgrim:
Latest commit fixes #61

ThePhilgrim:
@Akuli Would you please have another look at this?

Akuli:
I only read the diff of new changes since my previous review. Looks good!

PR #64, opened 2022-01-14
Use latin-1 internally, so that messages work with any encoding

### LINKED ISSUE #60: messages are assumed to be in UTF-8
Even though most of the world now uses UTF-8, I think the server still shouldn't crash if you use some different encoding.

To demonstrate this, I created a script that sends `hellö wörld` to the server with the latin-1 encoding instead of utf-8:

```python3
import socket
s = socket.socket()
s.connect(('localhost', 6667))
s.sendall("NICK x\r\n".encode("latin-1"))
s.sendall("USER x\r\n".encode("latin-1"))
s.sendall("JOIN #foo\r\n".encode("latin-1"))
s.sendall("PRIVMSG #foo :hellö world\r\n".encode("latin-1"))
```

This causes an error in mantatail:

```
Exception in thread Thread-4:
    self.run()
  File "/home/akuli/MantaTail/mantatail.py", line 101, in recv_loop
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf6 in position 4: invalid start byte
```

IRC servers should pass the messages from one client to another as they are, without attempting to interpret the bytes with a given encoding. The text of a PRIVMSG is not a string, as far as a correctly working server is concerned; it is just some bytes that gets transferred from one client to another, and it is the client's responsibility to interpret it as text.

Implementing this in mantatail is going to be annoying, because f-strings don't work with bytes. So `f"PRIVMSG {receiver} :{privmsg}"` must change to something quite different:

```python3
# If privmsg is bytes and everything else is strings:
f"PRIVMSG {receiver} :".encode() + privmsg

# If everything is bytes, both of these will work:
b"PRIVMSG %s :%s" % (receiver, privmsg)
b"PRIVMSG " + receiver + b" :" + privmsg
```

Akuli:
Fixes #60

### DIFF
--- a/commands.py
+++ b/commands.py
@@ -278,7 +278,7 @@ def error_unknown_command(user: mantatail.UserConnection, command: str) -> None:
 def error_not_registered() -> bytes:
     (not_registered_num, not_registered_info) = irc_responses.ERR_NOTREGISTERED
 
-    return bytes(f":mantatail {not_registered_num} * {not_registered_info}\r\n", encoding="utf-8")
+    return bytes(f":mantatail {not_registered_num} * {not_registered_info}\r\n", encoding="latin-1")
 
 
 def error_no_motd(user: mantatail.UserConnection) -> None:
@@ -291,7 +291,7 @@ def error_no_motd(user: mantatail.UserConnection) -> None:
 def error_nick_in_use(nick: str) -> bytes:
     (nick_in_use_num, nick_in_use_info) = irc_responses.ERR_NICKNAMEINUSE
 
-    return bytes(f":mantatail {nick_in_use_num} {nick} {nick_in_use_info}\r\n", encoding="utf-8")
+    return bytes(f":mantatail {nick_in_use_num} {nick} {nick_in_use_info}\r\n", encoding="latin-1")
 
 
 def error_no_such_nick_channel(user: mantatail.UserConnection, channel_or_nick: str) -> None:
--- a/mantatail.py
+++ b/mantatail.py
@@ -98,7 +98,7 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
                 else:
                     return  # go to "finally:"
 
-            decoded_message = request.decode("utf-8")
+            decoded_message = request.decode("latin-1")
             for line in split_on_new_line(decoded_message)[:-1]:
                 command, args = commands.parse_received_args(line)
                 command_lower = command.lower()
@@ -197,7 +197,7 @@ class UserConnection:
 
     def send_string_to_client(self, message: str, prefix: str) -> None:
         try:
-            message_as_bytes = bytes(f":{prefix} {message}\r\n", encoding="utf-8")
+            message_as_bytes = bytes(f":{prefix} {message}\r\n", encoding="latin-1")
 
             self.socket.sendall(message_as_bytes)
         except OSError:
--- a/tests/test_mantatail.py
+++ b/tests/test_mantatail.py
@@ -540,7 +540,6 @@ def test_invalid_utf8(user_alice, user_bob):
     while receive_line(user_bob) != b":mantatail 366 Bob #foo :End of /NAMES list.\r\n":
         pass
 
-    random_message = random.randbytes(100)
-    random_message = random_message.replace(b"\n", b"")
+    random_message = random.randbytes(100).replace(b"\n", b"")
     user_alice.sendall(b"PRIVMSG #foo :" + random_message + b"\r\n")
     assert receive_line(user_bob) == b":Alice!AliceUsr@127.0.0.1 PRIVMSG #foo :" + random_message + b"\r\n"
--- a/tests/test_mantatail.py
+++ b/tests/test_mantatail.py
@@ -1,3 +1,4 @@
+import os
 import pytest
 import random
 import socket
@@ -540,6 +541,6 @@ def test_invalid_utf8(user_alice, user_bob):
     while receive_line(user_bob) != b":mantatail 366 Bob #foo :End of /NAMES list.\r\n":
         pass
 
-    random_message = random.randbytes(100).replace(b"\n", b"")
+    random_message = os.urandom(100).replace(b"\n", b"")
     user_alice.sendall(b"PRIVMSG #foo :" + random_message + b"\r\n")
     assert receive_line(user_bob) == b":Alice!AliceUsr@127.0.0.1 PRIVMSG #foo :" + random_message + b"\r\n"
--- a/tests/test_mantatail.py
+++ b/tests/test_mantatail.py
@@ -528,3 +528,19 @@ def test_sudden_disconnect(run_server):
     nc.close()
 
     assert receive_line(nc2) == b":nc!nc@127.0.0.1 QUIT :Quit: (Remote host closed the connection)\r\n"
+
+
+def test_invalid_utf8(user_alice, user_bob):
+    user_alice.sendall(b"JOIN #foo\r\n")
+    time.sleep(0.1)
+    user_bob.sendall(b"JOIN #foo\r\n")
+
+    while receive_line(user_alice) != b":Bob!BobUsr@127.0.0.1 JOIN #foo\r\n":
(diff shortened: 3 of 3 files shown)


Akuli:
> I assume, the single-byte aspect is thus the main benefit here? But why is this actually beneficial?

The benefit comes from not only being single-byte, but also using all the bytes. This way, any byte is some valid character, and decoding cannot fail. I'll explain this in detail below.

First of all, a byte means an integer between 0 and 255:

```
>>> bytes([0, 7, 42, 123, 255])
b'\x00\x07*{\xff'
>>> bytes([256])
ValueError: bytes must be in range(0, 256)
```

Utf-8 supports many more than 256 characters, as you found out. It does it by sometimes using multiple bytes per character. For example, `ö` is two bytes, C3 B6, aka 195 182:

```
>>> "ö".encode("utf-8")
b'\xc3\xb6'
>>> list("ö".encode("utf-8"))
[195, 182]
>>> bytes([195, 182]).decode("utf-8")
'ö'
```

195 by itself is not valid UTF-8, because the rest is missing. UTF-8 is designed in such a way that it can tell if some bytes of a character are missing.

```
>>> bytes([195]).decode("utf-8")
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc3 in position 0: unexpected end of data
```

But with latin-1, this won't error. It produces one of the 256 characters that latin-1 supports.

```
>>> bytes([195]).decode("latin-1")
'Ã'
>>> bytes([195, 182]).decode("latin-1")
'Ã¶'
```

You can only get 256 characters this way. For example, latin-1 doesn't support `€`. Actually, here's a string contain all characters that latin-1 does support:

```
>>> valid_latin1_chars = bytes(list(range(256))).decode("latin-1")
>>> valid_latin1_chars
'\x00\x01\x02\x03\x04\x05\x06\x07\x08\t\n\x0b\x0c\r\x0e\x0f\x10\x11\x12\x13\x14\x15\x16\x17\x18\x19\x1a\x1b\x1c\x1d\x1e\x1f !"#$%&\'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~\x7f\x80\x81\x82\x83\x84\x85\x86\x87\x88\x89\x8a\x8b\x8c\x8d\x8e\x8f\x90\x91\x92\x93\x94\x95\x96\x97\x98\x99\x9a\x9b\x9c\x9d\x9e\x9f\xa0¡¢£¤¥¦§¨©ª«¬\xad®¯°±²³´µ¶·¸¹º»¼½¾¿ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖ×ØÙÚÛÜÝÞßàáâãäåæçèéêëìíîïðñòóôõö÷øùúûüýþÿ'
>>> len(valid_latin1_chars)
256
```

So if latin-1 supports **less** characters, how come it makes the server work **better** with funny messages?

Latin-1 accepts **anything**. There's no such thing as "invalid latin-1": any bytes are valid. (My test takes 100 random bytes with `os.urandom(100)` to ensure that this works.) For example, the two bytes `195, 182` from earlier turn into these two characters:

```
>>> bytes([195, 182]).decode("latin-1")
'Ã¶'
```

This is not `ö`, but it will go through the server as those characters, and come out as the same bytes:

```
>>> list('Ã¶'.encode("latin-1"))
[195, 182]
```

The IRC client can then interpret this as `ö`, if it has been configured to use UTF-8.

ThePhilgrim:
Wow, thank you for that really clear explanation. Seems very reasonable to do it this way.

One question remains in my mind, however:

You say 195 cannot be decoded by itself in utf-8. There are, however, other characters that are only one byte long.

Could the following issue ever arise?

Bytes [X, Y, 195] are decoded by the server to three characters, and are then supposed to be encoded again to go to a client using utf-8, however, utf-8 happens to encode (I just now realize that this is probably the opposite. The issue lies in Decoding, not Encoding.. but I'll finish my thought) X, Y together and then the character represented by 195 is left to encode by itself, which utf-8 cannot do, so it errors?


Lol, I wonder if this thought process makes sense when read by somebody else. But I guess that utf-8 can encode any character to a byte, but not decode parts of multi-byte characters by themselves.

Akuli:
> Bytes [X, Y, 195] are decoded by the server to three characters, and are then supposed to be encoded again to go to a client using utf-8

The server must be all latin-1. If it uses utf-8 anywhere at all, things will not work, because then bytes won't necessarily come out unchanged.

For example, let's see what happens if the server uses latin-1 when receiving data from clients, but incorrectly uses utf-8 when sending data to clients. I drew this picture after writing the explanation below, but I'm putting it first because I think it makes the explanation easier to follow:

```
BAD: Both clients use utf-8, server receives latin-1 but sends utf-8:

       utf-8                    latin-1    utf-8                                     utf-8
         |                         |          |                                        |
    Alice ------------------------>   server   --------------------------------------->   Bob
     1€      [49, 226, 130, 172]      1â\x82¬     [49, 195, 162, 194, 130, 194, 172]    1â\x82¬


GOOD: Both clients use utf-8, server uses latin-1 everywhere:

       utf-8                    latin-1    latin-1                    utf-8
         |                         |          |                         |
    Alice ------------------------>   server   ------------------------> Bob
     1€      [49, 226, 130, 172]      1â\x82¬     [49, 226, 130, 172]     1€

```

Alice types `1€` to her IRC client, and it uses UTF-8 to send it to the server. In UTF-8, this is four bytes, because `€` happens to be 3 bytes of utf-8.

```
>>> list("1€".encode("utf-8"))
[49, 226, 130, 172]
```

Then the server uses latin-1 to convert it to a string. This string is not `1€`, because `€` is not a valid character in latin-1. It is an internal string that represents the bytes that Alice sent, without interpreting it as UTF-8.

```
>>> internal_string_in_server = bytes([49, 226, 130, 172]).decode("latin-1")
>>> internal_string_in_server
'1â\x82¬'
```

Now, let's see what happens, **if the server incorrectly uses UTF-8** when sending this to Bob:

```
>>> server_to_bob = internal_string_in_server.encode("utf-8")
>>> list(server_to_bob)
[49, 195, 162, 194, 130, 194, 172]
```

Something has gone wrong! Bob is about to receive 7 bytes, even though Alice only sent 4 bytes. Let's see how this will show up in Bob's IRC client, if it uses UTF-8:

```
>>> server_to_bob.decode("utf-8")
'1â\x82¬'
```

So if the server uses UTF-8 when sending but latin-1 when receiving, Bob will see the message just like the server saw it: as some string that contains only latin-1 characters and likely isn't human-readable.

Now, let's see what happens if the server uses latin-1 when sending too, like it should:

```
>>> server_to_bob_2 = internal_string_in_server.encode("latin-1")
>>> list(server_to_bob_2)
[49, 226, 130, 172]
>>> server_to_bob_2.decode("utf-8")
'1€'
```

Akuli:
I think I should summarize why this works:
- Anything is valid latin-1 (because each byte simply corresponds with some character).
- `bytes.decode("latin-1").encode("latin-1")` will output the bytes as is: no matter what you send in, valid UTF-8 or not, it comes out unchanged. This is exactly what we want the server do.
- It is the clients' responsibility to use the same encoding (usually utf-8, but the server shouldn't mind if something else is used instead).
- ASCII characters just work*. Strings like `"PRIVMSG"` and `"JOIN"` don't need to be changed when we switch the internal encoding from utf-8 to latin-1.

* Specifically, UTF-8 and Latin-1 encode ASCII characters in the same way: `list("JOIN".encode("utf-8"))` and `list("JOIN".encode("latin-1"))` are both `[74, 79, 73, 78]`. So no matter how we convert between strings and bytes with utf-8 and latin-1, we can't mess up ASCII characters.

Akuli:
>But I guess that utf-8 can encode any character to a byte, but not decode parts of multi-byte characters by themselves.

This is correct :)

ThePhilgrim:
Thank you for clarifying this further. I can't find anything that needs further revision. Ready to merge.

PR #66, opened 2022-01-18
Docstrings

ThePhilgrim:
I have read a bunch of conventions on docstrings the last couple of days and the general opinion seems to be that: 

> Every module, class, method, and public function should have a docstring. These should allow a developer to go through and understand your program without having to read any code.

According to for example Google's guidelines for docstrings, they should include a summary of the element, what arguments it takes and what it returns.

Since many of my functions/methods are very obvious, I've opted for one-line solutions where reasonable.

I also like the thought of having well-documented code that also can be understood/read by beginners. As somebody who has a hard time reading code myself, this is something I wish more developers did in their projects.

With these thoughts in mind, I would ask you to consider if these docstrings

- Contain duplicate information
- Lack some information
- Some information is placed in the wrong place
- Some information doesn't make sense to include.


commands.py
```diff
@@ -1,3 +1,8 @@
+"""
+Contains handler functions that handle commands received from a client, as well as appropriate errors.
+
+All public functions must start with "handle_"
```
Akuli, line 4: This needs an explanation. Remember how hard it was for you to figure out how `http.server` works, before you knew about `getattr()`?

Explain the meaning of the `handle_` prefix as if the person reading the docstring was you before you learned about `getattr()`. People looking at the handler functions might not realize that the key to understanding them is the one `getattr()` line somewhere in the middle of 300 lines of `mantatail.py`.

commands.py
```diff
@@ -8,6 +13,17 @@
 
 ### Handlers
 def handle_join(state: mantatail.ServerState, user: mantatail.UserConnection, args: List[str]) -> None:
+    """
+    Handles clien't command to join a channel on the server.
+    Command format: "JOIN #foo"
+
+    If the channel already exists, the user is put to the appropriate channel's
+    Dict of connected users.
+    If the channel does not exist, the channel is created by instantiating mantatail.Channel.
```
Akuli, line 22: This contains lots of details that will get outdated as the code is refactored. Please try to explain it at a higher level, without mentioning specific dicts, class names, function names, variable names, and so on. If someone cares about those, they will have to read the rest of the function anyway.

```suggestion
    If the channel already exists, the user is added to the channel.
    If the channel does not exist, a new channel is created with the user as the founder.
```

This applies to a lot of other docstrings too. I'm not going to add a suggestion for each one :)

mantatail.py
```diff
@@ -1,3 +1,14 @@
+"""
+Represents the core of the server, with its main functionality and classes.
+
+All communication between server and client is encoded with latin-1.
+This ensures compatibility regardless of what encoding is used client-side.
+
+IRC Documentation:
+    - https://modern.ircdocs.horse/
+    - https://datatracker.ietf.org/doc/html/rfc1459
```
Akuli, line 9: Do we need the RFC for anything, when we have ircdocs.horse? I stopped looking at RFCs when you told me about ircdocs.horse.

mantatail.py
```diff
@@ -12,30 +23,44 @@
 
 
 class ServerState:
+    """
+    Represents the current state of the Server.
```
Akuli, line 27: If it's called `ServerState`, what would you expect? Please don't restate the obvious. See also https://thedailywtf.com/articles/documentation-on-your-contract

mantatail.py
```diff
@@ -56,28 +86,53 @@ def run_server_forever(self) -> None:
...
+    """
+    Ensures that the connection to a client is closed cleanly without errors.
+
+    The code is based on this blog post:
+    https://blog.netherlabs.nl/articles/2009/01/18/the-ultimate-so_linger-page-or-why-is-my-tcp-not-reliable
+
+    Args:
+        sock: Client socket
+
+    Raises:
```
Akuli, line 98: It doesn't raise anything. It handles and silences the error. That's an irrelevant detail to people using the function, so it belongs in a comment hidden inside the function, not in a docstring.

Akuli (commented):
I read a few of the docstrings, and spotted some overall problems that repeat in many other docstrings too. I only commented on a few example docstrings to give you an idea of what kinds of things need improving.

commands.py
ThePhilgrim: This makes sense, thanks!

mantatail.py
ThePhilgrim: I use RFC quite a lot still. Many of the IRC responses are not present in the horse docs. The horse docs also states that it's not complete.

ThePhilgrim:
@Akuli I updated the docstrings I could find that needed simplification, be less obvious, or contain less details/duplication of information.

commands.py
```diff
@@ -95,6 +123,15 @@ def handle_mode(state: mantatail.ServerState, user: mantatail.UserConnection, ar
 
 
 def handle_kick(state: mantatail.ServerState, user: mantatail.UserConnection, args: List[str]) -> None:
+    """
+    Handles client's command to kick a user from a channel.
```
Akuli, line 127: Every one of these starts with "Handles client's command to", or a misspelling of it. That's redundant and non-productive IMO. Can we explain it in the module's docstring instead of repeating in each function?

commands.py
```diff
@@ -128,12 +165,21 @@ def handle_kick(state: mantatail.ServerState, user: mantatail.UserConnection, ar
...
     user.send_que.put((None, None))
 
 
 def handle_privmsg(state: mantatail.ServerState, user: mantatail.UserConnection, args: List[str]) -> None:
-    # msg str
+    """
+    Handles client's command to send a message to a channel or a private message to a user.
+    Command format: "PRIVMSG #channel/user.nick :This is a message"
+
+    Private message's to users are handled in privmsg_to_user().
```
Akuli, line 181:
```suggestion
```

commands.py
```diff
@@ -163,6 +209,11 @@ def handle_privmsg(state: mantatail.ServerState, user: mantatail.UserConnection,
 
 
 def handle_pong(state: mantatail.ServerState, user: mantatail.UserConnection, args: List[str]) -> None:
+    """
+    Handles client's PONG resonse to a PING message sent from the server.
+
+    The PONG message notifies the server that the client still has an open connection to it.
+    """
```
Akuli, line 216: This is an example of a *perfect* docstring IMO :) It doesn't start with "Handles client's command to" or repeat anything else. But the best part is that it gives more context than what can be guessed from function name (server sends PING and client sends PONG, not vice versa). This gives the reader a very good idea of how ping/pong messaging works. It even explains the purpose of ping/ponging.

commands.py
```diff
@@ -269,41 +320,47 @@ def parse_received_args(msg: str) -> Tuple[str, List[str]]:
...
 def error_unknown_command(user: mantatail.UserConnection, command: str) -> None:
+    """Sent when server does not recognize a command user sent to server."""
     (unknown_cmd_num, unknown_cmd_info) = irc_responses.ERR_UNKNOWNCOMMAND
 
     message = f"{unknown_cmd_num} {command} {unknown_cmd_info}"
     user.send_que.put((message, "mantatail"))
 
 
 def error_not_registered(user: mantatail.UserConnection) -> None:
+    """Sent when a user sends a command before establishing a Nick and User."""
```
Akuli, line 331: You have typed NICK and USER to netcat so many times that you know very well what they do, but this could be very confusing for someone new to developing the server. What is "establishing a Nick and User" supposed to mean? Why are Nick and User capitalized?

You should at least mention that the NICK and USER you are referring to are commands, and that an IRC client must send them after connecting before it can use most other commands. Broader context like this is something that every new contributor lacks when looking at your code for the first time.

irc_responses.py
```diff
@@ -1,3 +1,17 @@
...
+Contains IRC-specific server replies that are sent to clients based on their actions.
+
+These include:
+    - Command Responses: Responses to certain commands, such as joining channel or setting channel mode.
+
+    - Error Replies: Responses to invalid actions by the user, for example,leaving a channel that
+    does not exist or not providing enough parameters in their commands.
+
+For a collection of all valid IRC responses:
+    - https://datatracker.ietf.org/doc/html/rfc1459#section-6
```
Akuli, line 11: I don't think this RFC contains **all** valid IRC responses. If I remember correctly, it's lacking many of the responses that ircdocs.horse has.

This makes me wonder if we should add links in docstrings at all, because they are already in the README.

mantatail.py
```diff
@@ -12,30 +23,47 @@
...
 
     def find_channel(self, channel_name: str) -> Channel:
+        """Looks for an existing channel and returns its channel object."""
         return self.channels[channel_name.lower()]
 
     def delete_user(self, nick: str) -> None:
+        """
+        Removes a user from all channels they are connected to,
+        thereafter removes user from connected users.
+        """
```
Akuli, line 52: This should probably mention that if you call this method, you must also make sure to send QUIT messages. We have had bugs where QUIT isn't sent in all cases, although the send queue does a good job of making sure they won't happen anymore.

mantatail.py
```diff
@@ -46,6 +74,11 @@ def __init__(self, port: int, motd_content: Optional[Dict[str, List[str]]]) -> N
         self.state = ServerState(motd_content)
 
     def run_server_forever(self) -> None:
+        """
+        Accepts incoming connections from clients.
+        Starts a receive loop on a separate thread,
+        listening for incoming commands from client.
+        """
```
Akuli, line 81: I don't like "listening" here, because in the context of sockets, it has a very specific meaning: being ready to accept new connections. That's not what the new threads do. How about something like this?

```suggestion
        """
        Accepts incoming connections from clients.
        Starts a separate thread to handle each connection.
        """
```

mantatail.py
```diff
@@ -56,28 +89,46 @@ def run_server_forever(self) -> None:
 
 
 def close_socket_cleanly(sock: socket.socket) -> None:
+    """Ensures that the connection to a client is closed cleanly without errors."""
```
Akuli, line 92:
```suggestion
    """Ensures that the connection to a client is closed cleanly without errors and with no data loss.

    Make sure that this is called before closing the socket.
    """
```

mantatail.py
```diff
@@ -56,28 +89,46 @@ def run_server_forever(self) -> None:
...
+    Note that netcat uses "\n" in place of "\r\n".
+
+    Ex: b"JOIN #foo\r\n"
+    Ex: b"PRIVMSG #foo :This is a message\r\n"
+
+    COMMAND is parsed to "handle_command", ex. "handle_join".
+    "getattr(commands, "handle_join")" checks if there is an appropriate handler function
+    in commands.py.
+    If there is, the handler function is called. Otherwise, an unknown command error is
+    sent to the client.
```
Akuli, line 124: This contains lots of technical details (getattr, error handling, the fact that the function name is saved to a string), and isn't easy to read (nested quotes, examples are a bit unclear). How about something like this:

```suggestion
    To handle a command FOO, a function named handle_foo() in commands.py is called.
    For example, "PRIVMSG #foo :this is a message\r\n" results in a call like this:

        commands.handle_privmsg(state, user, ["#foo", "this is a message"])
```

mantatail.py
```diff
@@ -56,28 +89,46 @@ def run_server_forever(self) -> None:
...
     sock.close()
 
 
 def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) -> None:
+    """
+    Instantiates UserConnection and listens for commands/messages from the client,
+    parses them and sends them to appropriate "handle_" function in "commands".
+
+    IRC Messages are formatted "bytes(COMMAND args\r\n)"
+    Note that netcat uses "\n" in place of "\r\n".
```
Akuli, line 115:
```suggestion
    Most IRC clients use "\r\n" line endings, but "\n" is accepted as well (used by e.g. netcat).
```

mantatail.py
```diff
@@ -133,6 +184,19 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
...
+    Represents the connection between server & client.
+    Starts a Queue on a separate thread on which the client receives all messages from the server.
+
+    Format examples:
+    - Nick: Alice
+    - User message: AliceUsr 0 * Alice's Real Name
+    - Username: AliceUsr
+    - User Mask Alice!AliceUsr@127.0.0.1 (Nick!Username@Host)
+
+    All references to UserConnection (lists, dicts, etc.) are based on Nick.
```
Akuli, line 197: This again expresses a simple thing using lots of complicated technical details.

```suggestion
    Usually the nick is used when referring to the user.
```

mantatail.py
```diff
@@ -133,6 +184,19 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
 
 
 class UserConnection:
+    """
+    Represents the connection between server & client.
+    Starts a Queue on a separate thread on which the client receives all messages from the server.
+
+    Format examples:
+    - Nick: Alice
+    - User message: AliceUsr 0 * Alice's Real Name
```
Akuli, line 193:
```suggestion
    - User message: AliceUsr 0 * Alice's Real Name
```

mantatail.py
```diff
@@ -133,6 +184,19 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
 
 
 class UserConnection:
+    """
+    Represents the connection between server & client.
+    Starts a Queue on a separate thread on which the client receives all messages from the server.
```
Akuli, line 189:
```suggestion

    A send queue and a separate thread are used for sending.
    This helps with error handling, and even if someone has a slow internet connection,
    other people don't have to wait when a message is sent to several users with a loop.
```

mantatail.py
```diff
@@ -149,13 +213,21 @@ def __init__(self, state: ServerState, host: str, socket: socket.socket):
...
         return f"{self.nick}!{self.user_name}@{self.host}"
 
     def send_queue_thread(self) -> None:
+        """
+        Queue on which the client receives messages from server.
+
+        All messages are a tuple formatted as (message, prefix).
+        Prefixes are either ":mantatail" or "sender.user_mask"
+
+        A Tuple containing (None, None) indicates a QUIT and closes the connection to the client.
```
Akuli, line 226: These details about how the send queue works belong in the docstring of the class. People who use the class shouldn't have to read all methods just to understand the send queue.

mantatail.py
```diff
@@ -211,21 +288,41 @@ def send_string_to_client(self, message: str, prefix: Optional[str]) -> None:
...
+        """
+        Starts a timer on a separate thread that, when finished, sends a PING message to the client
+        to establish that the client still has an open connection to the server.
+        """
         self.ping_timer = threading.Timer(TIMER_SECONDS, self.queue_ping_message)
         self.ping_timer.start()
 
     def queue_ping_message(self) -> None:
+        """
+        Puts PING message in the client's Queue, and starts a new timer waiting for the
```
Akuli, line 300: We should probably refer to the queue as "the send queue" or "the client's send queue", instead of "Queue". Just calling it Queue seems ambiguous and misleading to me.

mantatail.py
```diff
@@ -211,21 +288,41 @@ def send_string_to_client(self, message: str, prefix: Optional[str]) -> None:
...
+        Sends ":mantatail PING :mantatail"
+        Expected response: ":Alice!AliceUsr@127.0.0.1 PONG :mantatail"
+        """
         self.send_que.put(("PING :mantatail", "mantatail"))
         threading.Timer(5, self.assert_pong_received).start()
 
     def assert_pong_received(self) -> None:
+        """
+        Asserts if the client has sent an appropriate
+        PONG response to the server's PING message.
```
Akuli, line 313: What is  this supposed to mean? Asserting means you get an error if a condition is false. That's not happening here. Can you clarify this a bit, and maybe rename the method too?

mantatail.py
```diff
@@ -211,21 +288,41 @@ def send_string_to_client(self, message: str, prefix: Optional[str]) -> None:
...
+        If no PONG response has been received, the server closes the connection to the client.
+        """
         if not self.pong_received:
             self.send_que.put((None, None))
         else:
             self.pong_received = False
 
 
 class Channel:
+    """Creates a channel on the server"""
```
Akuli, line 324: This states the obvious. It's a class `Channel` in an IRC server project. It's fine to not have a docstring for everything, especially if there's nothing to say.

mantatail.py
```diff
@@ -237,29 +334,38 @@ def __init__(self, channel_name: str, user: UserConnection) -> None:
...
+        Thereafter removes the kicked user from the channel.
+        """
         for usr in self.users:
             usr.send_que.put((message, kicker.get_user_mask()))
 
         self.users.discard(user_to_kick)
 
 
 def split_on_new_line(string: str) -> List[str]:
+    """Splits a message received by a client on "\r\n" (IRC) or "\n" (Netcat)."""
```
Akuli, line 360: I don't like `(IRC)` and `(Netcat)`

```suggestion
    """Splits a message received by a client on "\r\n" (most IRC clients) or "\n" (e.g. Netcat)."""
```

mantatail.py
```diff
@@ -237,29 +334,38 @@ def __init__(self, channel_name: str, user: UserConnection) -> None:
...
     if string.endswith("\r\n"):
         return string.split("\r\n")
     else:
         return string.split("\n")
 
 
 def get_motd_content_from_json() -> Optional[Dict[str, List[str]]]:
+    """Fetches the server's Message of the Day."""
```
Akuli, line 368:
```suggestion
    """Loads the Message of the Day file.

    Returns None if the file is not found.
    """
```

Akuli (commented):
You seem to have mostly understood the overall idea now, so I reviewed the docstrings in detail. Sometimes you're spot on, but there were still a few docstrings that contained too many technical details that confuse the reader and will get outdated easily.

In general, try to make the docstrings readable for someone very new to mantatail. Imagine that you have never written NICK and USER to netcat, you don't understand the problems of not having a send queue, you have never seen `getattr()` anywhere before, and so on.

commands.py
Akuli: And even with all that, it's concise and to the point.

mantatail.py
Akuli: I didn't mean to leave a suggestion on this line :)

ThePhilgrim:
Issue #68 will be fixed in a separate PR

commands.py
```diff
@@ -1,3 +1,19 @@
+"""
+Contains handler functions that handle commands received from a client, as well as appropriate errors.
+
+Each command can include:
+    - Tags: Optional metadata on a message, starting with '@'.
+    - Source: Optional note of where the message came from, starting with ':'.
+        * This is usually the server name or the user mask
+    - Command: The specific command this message represents.
+    - Parameters: Optional data relevant to this specific command – a series of values
+        separated by one or more spaces. Parameters have different meanings for every single message.
```
Akuli, line 10: This feels very abstract and difficult to read to me. Can you show an example of a full received line and point out what exactly is a tag in it, what is a source, etc?

commands.py
```diff
@@ -84,6 +117,15 @@ def handle_part(state: mantatail.ServerState, user: mantatail.UserConnection, ar
...
 
 def handle_mode(state: mantatail.ServerState, user: mantatail.UserConnection, args: List[str]) -> None:
+    """
+    Command format: "MODE #channel/user.nick +/-flag <args>"
+
+    Sets a user/channel mode.
+
+    Ex:
+        - User mode "+i" makes user invisible
+        - Channel mode "+i" makes channel invite-only.
```
Akuli, line 127: You should maybe note that neither of these is supported yet.

commands.py
```diff
@@ -254,6 +336,18 @@ def process_user_modes() -> None:
 
 
 def parse_received_args(msg: str) -> Tuple[str, List[str]]:
+    """
+    Parses the user command by separating the command (e.g "join", "privmsg", etc.) from the
+    arguments.
+
+    If the optional "data portion" of the parameters contains one or more spaces and starts with ':',
```
Akuli, line 343: Is "data portion" the name used in ircdocs.horse? If not, does ircdocs.horse have a name for this?

commands.py
```diff
@@ -313,58 +416,85 @@ def error_not_on_channel(user: mantatail.UserConnection, channel_name: str) -> N
 def error_user_not_in_channel(
     user: mantatail.UserConnection, target_usr: mantatail.UserConnection, channel: mantatail.Channel
 ) -> None:
+    """
+    Sent when a user sends a channel-specific command with a user as an argument,
+    and this user is connected to the server but not to the channel.
```
Akuli, line 421: You connect to a server and join a channel. You don't connect to a channel.

```suggestion
    and this user is connected to the server but has not joined the channel.
```

mantatail.py
```diff
@@ -12,30 +19,50 @@
 
 
 class ServerState:
+    """Keeps track of existing channels & connected users."""
+
     def __init__(self, motd_content: Optional[Dict[str, List[str]]]) -> None:
+        """
+        Attributes:
+            lock: Locks the state of the server to avoid modifications to iterables during iteration.
```
Akuli, line 27:
```suggestion
            lock: Ensure that the state isn't used simultaneously in multiple places.
```

mantatail.py
```diff
@@ -12,30 +19,50 @@
...
+            channels: The currently existing channels on the server.
+            connected_users: The currently connected users on the server.
```
Akuli, line 29: These seem to state the obvious, at least to me. If `ServerState` has an attribute `channels`, what else could it be?

mantatail.py
```diff
@@ -56,28 +87,49 @@ def run_server_forever(self) -> None:
...
+        # - Client was already disconnected.
+        # - Probably something else too that I didn't think of...
         pass
 
     sock.close()
 
 
 def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) -> None:
+    """
+    Instantiates UserConnection and listens for commands/messages from the client,
```
Akuli, line 113: - Instantiating UserConnection is an implementation detail. It doesn't belong in the docstring.
- This is not listening; that means accepting new client connections. It is receiving

```suggestion
    Receives commands/messages from the client,
```

mantatail.py
```diff
@@ -200,6 +276,10 @@ def queue_quit_message_for_other_users(self) -> None:
             receiver.send_que.put((message, self.get_user_mask()))
 
     def send_string_to_client(self, message: str, prefix: Optional[str]) -> None:
+        """
+        Formats a message taken from the user's send queue, converts it to
+        bytes (encoded with latin-1) and sends it to the client.
+        """
```
Akuli, line 282: While this explains exactly what the method does, it isn't very useful. Someone working on Mantatail will likely find this, think that they are supposed to use it for everything, and never learn about the send queue.

```suggestion
        """
        Send a string to the client, without using the send queue.

        In most cases, you should put a message to the send queue instead of using this method.
        """
```

Akuli (commented):
It's getting pretty good! I left a few final nits. Feel free to merge when fixed.

commands.py
ThePhilgrim: Removing `Tags` since it's currently not used in mantatail, and seems like something that is not commonly used.

commands.py
ThePhilgrim: > Parameters (or ‘params’) are extra pieces of information added to the end of a message. These parameters generally make up the ‘data’ portion of the message. What specific parameters mean changes for every single message.

Taken straight from IRC Horse

Akuli: Maybe just say "If there is `:` following a space, everything after the `:` is interpreted as a single parameter" or something like that?

PR #67, opened 2022-01-18
Fix user prefix not showing on self when joining channel

### LINKED ISSUE #62: Names list (353) does not include founder mode
On libera, if my nickname is `asdf2` and I join a channel that doesn't exist, the names list includes me:

```
JOIN #akuli-testing
:asdf2!~asdf2@82-203-163-9.bb.dnainternet.fi JOIN #akuli-testing
:strontium.libera.chat MODE #akuli-testing +Cnst
:strontium.libera.chat 353 asdf2 @ #akuli-testing :@asdf2
:strontium.libera.chat 366 asdf2 #akuli-testing :End of /NAMES list.
```

Ignoring the noise, `:@asdf2` is the list of users on this channel, and I am an operator because I have `@` in front of my nickname. The `@` character is also repeated just before the channel name.

With mantatail, I get:

```
JOIN #akuli-testing
:asdf2!asdf2@127.0.0.1 JOIN #akuli-testing
:mantatail 353 asdf2 = #akuli-testing :asdf2 
:mantatail 366 asdf2 #akuli-testing :End of /NAMES list.
```

Note that there is `=`, which presumably means I am nothing special: not a founder, not an operator, not anything else. I also don't have anything prefixed in front of my name. But if another user (named `x`) joins, their names list indicates that I am the founder, because I then get to have `~` in front of my name:

```
:mantatail 353 x = #akuli-testing :x ~asdf2
```

ThePhilgrim:
Fixes #62

### DIFF
--- a/commands.py
+++ b/commands.py
@@ -42,8 +42,14 @@ def handle_join(state: mantatail.ServerState, user: mantatail.UserConnection, ar
                 usr.send_que.put((message, user.get_user_mask()))
 
             # TODO: Implement topic functionality for existing channels & MODE for new ones
+            if channel.is_founder(user):
+                own_prefix = "~"
+            elif channel.is_operator(user):
+                own_prefix = "@"
+            else:
+                own_prefix = ""
 
-            message = f"353 {user.nick} = {channel_name} :{user.nick} {channel_users_str.strip()}"
+            message = f"353 {user.nick} = {channel_name} :{own_prefix}{user.nick} {channel_users_str.strip()}"
             user.send_que.put((message, "mantatail"))
 
             message = f"366 {user.nick} {channel_name} :End of /NAMES list."
--- a/commands.py
+++ b/commands.py
@@ -37,10 +37,7 @@ def handle_join(state: mantatail.ServerState, user: mantatail.UserConnection, ar
 
             # TODO: Implement topic functionality for existing channels & MODE for new ones
 
-            message = f"353 {user.nick} = {channel_name} :{user.get_nick_with_prefix(channel)}"
-
-            if channel_users_str:
-                message += f" {channel_users_str.strip()}"
+            message = f"353 {user.nick} = {channel_name} :{user.get_nick_with_prefix(channel)}{channel_users_str}"
 
             user.send_que.put((message, "mantatail"))
 
--- a/commands.py
+++ b/commands.py
@@ -27,13 +27,7 @@ def handle_join(state: mantatail.ServerState, user: mantatail.UserConnection, ar
         if user not in channel.users:
             channel_users_str = ""
             for usr in channel.users:
-                if usr.user_name == channel.founder:
-                    nick = f"~{usr.nick}"
-                elif channel.is_operator(usr):
-                    nick = f"@{usr.nick}"
-                else:
-                    nick = usr.nick
-                channel_users_str += f" {nick}"
+                channel_users_str += f" {usr.get_nick_with_prefix(channel)}"
 
             channel.users.add(user)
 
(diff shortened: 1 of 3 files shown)


mantatail.py
```diff
@@ -242,6 +242,15 @@ def set_operator(self, user: UserConnection) -> None:
     def remove_operator(self, user: UserConnection) -> None:
         self.operators.discard(user.nick.lower())
 
+    def is_founder(self, user: UserConnection) -> bool:
+        """
+        Takes a UserConnection object as an argument and checks if the user
+        is the channel founder.
+
+        Returns a Boolean.
+        """
```
Akuli, line 251: I don't like repeating what the type annotations already say. You should assume that the reader knows how to read type annotations, because your code is full of them anyway.

```suggestion
        """Checks if the user is the channel founder."""
```

commands.py
```diff
@@ -42,8 +42,14 @@ def handle_join(state: mantatail.ServerState, user: mantatail.UserConnection, ar
...
 
             # TODO: Implement topic functionality for existing channels & MODE for new ones
-
-            message = f"353 {user.nick} = {channel_name} :{user.nick} {channel_users_str.strip()}"
+            if channel.is_founder(user):
+                own_prefix = "~"
+            elif channel.is_operator(user):
+                own_prefix = "@"
+            else:
+                own_prefix = ""
```
Akuli, line 50: This logic is now duplicated in two places. Can we put it to a new method, instead of basically copy/pasting it?

tests/test_mantatail.py
```diff
@@ -331,9 +331,15 @@ def test_channel_owner(user_alice, user_bob):
...
 
-def test_operator_prefix(user_alice, user_bob, user_charlie):
+def test_founder_and_operator_prefix(user_alice, user_bob, user_charlie):
     user_alice.sendall(b"JOIN #foo\r\n")
-    time.sleep(0.1)
+
+    while True:
+        received = receive_line(user_alice)
+        if b"353" in received:
+            assert compare_if_word_match_in_any_order(received, b":mantatail 353 Alice = #foo :~Alice\r\n")
```
Akuli, line 340: `assert received == b"..."` should work here, because there's only one user.

ThePhilgrim:
@Akuli I think this PR is ready for a final check

tests/test_mantatail.py
```diff
@@ -331,9 +331,12 @@ def test_channel_owner(user_alice, user_bob):
             break
 
 
-def test_operator_prefix(user_alice, user_bob, user_charlie):
+def test_founder_and_operator_prefix(user_alice, user_bob, user_charlie):
     user_alice.sendall(b"JOIN #foo\r\n")
-    time.sleep(0.1)
+    receive_line(user_alice)  # JOIN message from server
+
+    assert compare_if_word_match_in_any_order(receive_line(user_alice), b":mantatail 353 Alice = #foo :~Alice\r\n")
```
Akuli, line 338: There's only one nick, so the order should be exactly as in the tests and `compare_if_word_match_in_any_order` isn't needed:

```suggestion
    assert receive_line(user_alice) == b":mantatail 353 Alice = #foo :~Alice\r\n"
```

Akuli:
I rewrote a test slightly above, and I think I found a bug. The list of nicknames sometimes arrives with an unnecessary space at the end.

commands.py
```diff
@@ -37,9 +37,11 @@ def handle_join(state: mantatail.ServerState, user: mantatail.UserConnection, ar
 
             # TODO: Implement topic functionality for existing channels & MODE for new ones
 
-            message = (
-                f"353 {user.nick} = {channel_name} :{user.get_nick_with_prefix(channel)} {channel_users_str.strip()}"
-            )
+            message = f"353 {user.nick} = {channel_name} :{user.get_nick_with_prefix(channel)}"
+
+            if channel_users_str:
+                message += f" {channel_users_str.strip()}"
```
Akuli, line 43: So... we remove the space in the beginning with `.strip()`, and then add a new space with an f-string?

After simplifying this a bit, you might realize that it can be simplified even more :)

Akuli (approved):

PR #71, opened 2022-01-21
Add logic for sending private message to user

ThePhilgrim:
An error arose in `find_user()` and `find_channel()` when adding the test for private message error.
That's why they are changed.

Everything else seems to work properly. See if any mantaray tests that used to fail now pass


Akuli:
Nice! I'll upgrade the MantaTail used in tests once this PR is merged.

Akuli (approved):

Issue #77, opened 2022-01-21
AttributeError: 'UserConnection' object has no attribute 'nick'


Akuli:
Start MantaTail, connect to it with netcat, and then immediately disconnect without send anything. You will get error

```
Exception in thread Thread-71:
Traceback (most recent call last):
  File "/usr/lib/python3.9/threading.py", line 954, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.9/threading.py", line 892, in run
    self._target(*self._args, **self._kwargs)
  File "/home/akuli/MantaTail/mantatail.py", line 253, in send_queue_thread
    self.queue_quit_message_for_other_users()
  File "/home/akuli/MantaTail/mantatail.py", line 289, in queue_quit_message_for_other_users
    if channel.is_operator(self):
  File "/home/akuli/MantaTail/mantatail.py", line 376, in is_operator
    return user.nick.lower() in self.operators
AttributeError: 'UserConnection' object has no attribute 'nick'
```

Akuli:
Actually, reproducing the bug needs a different user to join some channel:

1. In one terminal, send `NICK x`, `USER x`, `JOIN #foo`
2. In a second terminal, close netcat without sending anything.

Akuli:
Already fixed in #79, but it would be good to add a test.

PR #79, opened 2022-01-21
Nickchanges

### LINKED ISSUE #74: IndexError
Connect to mantatail with netcat and type `NICK` with no parameters.

```
Mantatail running (127.0.0.1:6667)
Exception in thread Thread-1:
    self.run()
  File "/home/akuli/MantaTail/mantatail.py", line 163, in recv_loop
    if args[0].lower() in state.connected_users.keys():
```

ThePhilgrim:
Fixes #74


commands.py
```diff
@@ -136,6 +136,43 @@ def handle_mode(state: mantatail.ServerState, user: mantatail.UserConnection, ar
...
+            for receiver in receivers:
+                receiver.send_que.put((message, user.get_user_mask()))
+
+            # User doesn't get NICK message if they change their nicks before sending USER command
+            if user.user_message:
+                user.send_que.put((message, user.get_user_mask()))
+
+            del state.connected_users[user.nick.lower()]
+            user.nick = new_nick
+            state.connected_users[user.nick.lower()] = user
```
Akuli, line 173: Create a test where an operator changes nick and then tries to KICK someone. You will notice a problem.

commands.py
```diff
@@ -136,6 +136,43 @@ def handle_mode(state: mantatail.ServerState, user: mantatail.UserConnection, ar
...
+    Changes the user's nickname if they already have one.
+    """
+    nick_regex = r"[a-zA-Z|\\_\[\]{}^`-][a-zA-Z0-9|\\_\[\]{}^`-]{,15}"
+
+    if not args:
+        error_no_nickname_given(user)
+        return
+
+    new_nick = args[0]
+    if not re.match(nick_regex, new_nick):
```
Akuli, line 151: `re.match` is a terribly named function. For example, `re.match("hello", "hello world")` matches! A much better name for it would be `match_beginning` or `match_prefix` or something.

```suggestion
    if not re.fullmatch(nick_regex, new_nick):
```

Akuli:
If this fixes #74, please add a test that fails on main because of #74.

commands.py
ThePhilgrim: Ah, well spotted! I see why this would become an issue

commands.py
ThePhilgrim: Oh, I thought `match` was a method in the `re` module. I didn't realize we could name these functions ourselves.

commands.py
Akuli: This is why I disliked storing nickname strings all over the server state :)

commands.py
Akuli: It is in the `re` module, and you can't name it yourself. I'm saying that whoever created the `re` module named it badly.

commands.py
ThePhilgrim: This is fixed in most recent commit (yet to be pushed)

Akuli:
Can you also add tests where:
- `Alice` changes their nick to `ALICE`
- `Bob` changes their nick to `Bob` (i.e. doesn't actually change the nick)

I'm not sure what these should do, but you can try with libera.

ThePhilgrim:
> Can you also add tests where:
> 
> * `Alice` changes their nick to `ALICE`
> * `Bob` changes their nick to `Bob` (i.e. doesn't actually change the nick)
> 
> I'm not sure what these should do, but you can try with libera.

Done!

According to Libera, changing nick from `Alice` to `ALICE` should just change the nick as usual.
Changing nick from `Alice` to `Alice` is just silently ignored.

These situations are both present in the test update.

mantatail.py
```diff
@@ -348,26 +349,26 @@ def __init__(self, channel_name: str, user: UserConnection) -> None:
...
 
     def set_operator(self, user: UserConnection) -> None:
         """Adds a user to the channel's operators."""
-        self.operators.add(user.nick.lower())
+        self.operators.add(user)
 
     def remove_operator(self, user: UserConnection) -> None:
         """Removes a user from the channel's operators."""
-        self.operators.discard(user.nick.lower())
+        self.operators.discard(user)
```
Akuli, line 363: These methods are now so simple that they should be removed IMO. I don't see how `channel.set_operator(user)` is any better than `channel.operators.add(user)`. We originally added the methods because `channel.operators.add(user.nick.lower())` is ugly and fragile.

Akuli (approved):
Looks good! Just one idea for cleanup that can be done in a separate PR, if you prefer.

Akuli:
I believe this fixes #77, because the error in #77 comes from `is_operator()`. Can you add a test for that too?

ThePhilgrim:
> I believe this fixes #77, because the error in #77 comes from `is_operator()`. Can you add a test for that too?

@Akuli This has a more solid fix in #86
I'll add any tests related to this in that PR.

Issue #81, opened 2022-01-21
two problems with joining new channel and kicking self


Akuli:
![x](https://user-images.githubusercontent.com/18505570/150609994-6b8d5779-5c85-4075-877b-e533dda880b5.png)

There is never anybody else on `#foo` than the user `x`, so if I disconnect and reconnect, the behavior should be the same each time. But it isn't. Somehow the existence of `#foo` is remembered even if everyone disconnects from `#foo`.

Also, if I am a founder, how is it possible that I am not an operator ????!

PR #86, opened 2022-01-22
Bugfixes

### LINKED ISSUE #69: AttributeError: 'UserConnection' object has no attribute 'nick'
I don't know yet when and why this happens, but it can happen. Here's an error message, taken from https://github.com/Akuli/mantaray/runs/4889957324?check_suite_focus=true

```
  File "/opt/hostedtoolcache/Python/3.10.1/x64/lib/python3.10/threading.py", line 1009, in _bootstrap_inner
    self.run()
  File "/opt/hostedtoolcache/Python/3.10.1/x64/lib/python3.10/threading.py", line 946, in run
  File "/home/runner/work/mantaray/mantaray/mantatail/mantatail.py", line 173, in send_queue_thread
  File "/home/runner/work/mantaray/mantaray/mantatail/mantatail.py", line 208, in queue_quit_message_for_other_users
    if channel.is_operator(self):
  File "/home/runner/work/mantaray/mantaray/mantatail/mantatail.py", line 262, in is_operator
    return user.nick.lower() in self.operators
```

### LINKED ISSUE #76: "MODE +o #foo y" fails silently
I accidentally typed `MODE +o #foo y` instead of `MODE #foo +o y`. Surprisingly, I didn't get any error message. It just silently did nothing.

### LINKED ISSUE #77: AttributeError: 'UserConnection' object has no attribute 'nick'

### LINKED ISSUE #83: response 482 is wrong
libera:

```
KICK #mantatest xyz
:molybdenum.libera.chat 482 asdfff #mantatest :You're not a channel operator
```

mantatail:

```
KICK #asdf x
:mantatail 482 #asdf :You're not channel operator
```

Note that between 482 and channel name, libera puts nickname of the current user.

### LINKED ISSUE #84: Nick `*` in errors before registering
When Libera sends an error before the client has registered, the client part of the error message is substituted with `*`.
Skärmavbild 2022-01-22 kl  11 27 06

This could be solved either by setting user.nick to `*` at instantiation and then do checks like `if user.nick != "*"`instead of `hasattr`.

The other way is to check in the error messages `if hashattr(user, "nick")` and if not, set client part to `*`

@Akuli What would you say is the best solution here?

### LINKED ISSUE #88: RPL_CHANNELMODEIS is broken
https://github.com/ThePhilgrim/MantaTail/blob/0bb8cab2493930f9caca0acb8d734fc51f18c7a3/commands.py#L332-L349

We never get to `if len(args) == 1:`, because if there is only one argument, `args[1][0]` will be an IndexError.

```
>>> args = ["#foo"]
>>> len(args)
1
>>> args[1][0]
```

This means that no matter what you do, you can't get the server to send `RPL_CHANNELMODEIS`.

ThePhilgrim:
Currently:

- Fixes #76 
- Fixes #88


ThePhilgrim:
@Akuli I'm trying to figure out a "bug" that I'm running into in the tests. When Alice `kick` Bob, Alice receives first `:Alice!AliceUsr@127.0.0.1 KICK #foo Bob :Bob\r\n'`, and then she receives `b'\r\n'`

I really can't figure out where the empty bytes come from. It seems like this is not received when using Mantaray's --verbose to kick.

Test:
Skärmavbild 2022-01-23 kl  12 48 28

Output:
Skärmavbild 2022-01-23 kl  12 50 59

Any ideas?

Akuli:
`handle_kick` puts a message to send queue with \r\n. Most other `handle_foo` functions don't.

ThePhilgrim:
@Akuli Thank you!

Another question:

The user is in this case the channel founder. With the check below, the code does not run (channel.is_founder returns `True`)
```
if not channel.is_founder(user):
        print("CURRENT: ", user.user_name)
        print("FOUNDER: ", channel.founder)
        print(user.user_name == channel.founder)
        error_no_operator_privileges(user, channel)
        return
```

However, when I add `or`, it suddenly makes the code run:

```
if not channel.is_founder(user) or (user not in channel.operators):
        print("CURRENT: ", user.user_name)
        print("FOUNDER: ", channel.founder)
        print(user.user_name == channel.founder)
        error_no_operator_privileges(user, channel)
        return
```

Note: I tried both `if not channel.is_founder(user) or user not in channel.operators`, `if not channel.is_founder(user) or (user not in channel.operators)`, & `if (not channel.is_founder(user)) or (user not in channel.operators)`

Why is this if check suddenly breaking?

Akuli:
You have a mess of logic, but what problem are you actually trying to solve?

I see Mantatail's founder stuff as a "good enough for now" solution. It doesn't support multiple founders, and anyone can become a founder by just registering with the same `USER` line that the founder used. I believe founders on Libera are more like operators, just a flag that you can enable/disable user-specifically on each channel.

ThePhilgrim:
@Akuli I don't see why this is a mess of logic? Apart from the prints it's just sending a single error message.

Currently, if `user A` and `user B` are on channel `#foo` -> `user A` `KICK` `user A` -> `user A QUIT` -> `user A` connects and `JOIN #foo` -> `user A` `KICK` `user B`, there is error message `You are not channel operator` (issue #81 )

The current code is:
```
if user not in channel.operators:
        error_no_operator_privileges(user, channel)
        return
```

Instead, I want to check if A) user is in channel.operator, or B) user.user_name is the same as channel.founder. That means that even if the founder leaves a channel, they will still be the channel founder, and be able to OP themselves or KICK other users. That's why I need the `if not channel.is_founder(user) or (user not in channel.operators)` above, and don't understand why it breaks the logic.

ThePhilgrim:
Now fixes #81

ThePhilgrim:
The logic for commit f2f8264 is based on Libera's logic.

If I send `MODE ThePhilgrim #foo +o`, Libera returns `403 No such channel` if there is no user `ThePhilgrim` connected.
If there is a user `ThePhilgrim` connected, Libera returns `502 Can't change mode for other users`.

In the same way, if I send `MODE +o #foo ThePhilgrim`, Libera returns `403`.

Since User Modes are still not implemented in Mantatail, I have not researched if `502` applies for any mode when trying to set them on another user. Therefore I decided to interpret `args[0]` as the channel for all cases for now.

I prepared for this logic for the future by implementing the `if target_usr != user` check.

ThePhilgrim:
Now the user can properly check what the channel modes are by sending command `MODE #channel`. There are, however, currently no channel modes that are added to that list in mantatail (+o doesn't seem to get added as a channel mode in Libera).

Akuli:
Ummm... do you really think I will review a pull request that fixes 7 different bugs at once? :D

Please split it up into smaller pull requests. You can use `git cherry-pick`.

ThePhilgrim:
Ah, I didn't realize it was more difficult to review a PR like this. I'll see if I can understand how to cherry pick several commits into a new branch and split this into smaller PRs

Akuli:
Everything that this PR adds is already on main. I looked at `git diff bugfixes main` with both branches pulled.

PR #87, opened 2022-01-22
Delete now-unnecessary set/remove/is_operator methods

Akuli:


mantatail.py
```diff
@@ -352,24 +351,12 @@ def __init__(self, channel_name: str, user: UserConnection) -> None:
...
-
-    def set_operator(self, user: UserConnection) -> None:
-        """Adds a user to the channel's operators."""
         self.operators.add(user)
 
-    def remove_operator(self, user: UserConnection) -> None:
-        """Removes a user from the channel's operators."""
-        self.operators.discard(user)
-
     def is_founder(self, user: UserConnection) -> bool:
```
ThePhilgrim, line 356: Wouldn't it, in this case, be reasonable to delete `is_founder()` as well and do a `if user.user_name == channel.founder` check? The level of complexity seems to be similar.

Akuli: I can do that if you want. It is about the same complexity, but `is_founder()` prevents you from accidentally using `nick` when you should be using `user_name`.

ThePhilgrim: Okay, that makes sense to me. With that in mind, I will merge this. Thanks!

PR #89, opened 2022-01-23
Issue hunt1

### LINKED ISSUE #77: AttributeError: 'UserConnection' object has no attribute 'nick'

### LINKED ISSUE #83: response 482 is wrong

### LINKED ISSUE #84: Nick `*` in errors before registering

ThePhilgrim:
Fixes #83
Fixes #84
Fixes #77

### DIFF
--- a/mantatail.py
+++ b/mantatail.py
@@ -206,7 +206,7 @@ class UserConnection:
         self.state = state
         self.socket = socket
         self.host = host
-        self.nick: str = "*"
+        self.nick = "*"
         self.user_message: Optional[List[str]] = None  # Ex. AliceUsr 0 * Alice
         self.user_name: Optional[str] = None  # Ex. AliceUsr
         self.send_que: queue.Queue[Tuple[str, str] | Tuple[None, None]] = queue.Queue()
--- a/mantatail.py
+++ b/mantatail.py
@@ -149,7 +149,7 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
                 command_lower = command.lower()
                 parsed_command = "handle_" + command_lower
 
-                if not hasattr(user, "nick") or not user.user_message:
+                if user.nick == "*" or not user.user_message:
                     if command_lower == "user":
                         user.user_message = args
                         user.user_name = args[0]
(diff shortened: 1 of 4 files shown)


mantatail.py
```diff
@@ -202,14 +202,11 @@ class UserConnection:
...
 
-    # self.nick is defined in recv_loop()
-    # It is not set in __init__ to keep mypy happy.
-    nick: str  # Nick is shown in user lists etc, user_name is not
-
     def __init__(self, state: ServerState, host: str, socket: socket.socket):
         self.state = state
         self.socket = socket
         self.host = host
+        self.nick: str = "*"
```
Akuli, line 209: Mypy can now correctly guess that `nick` is a string. There's no need to tell it to mypy manually.

```suggestion
        self.nick = "*"
```

Akuli (commented):
Just a small nit:

Akuli:
I applied my suggestion from review and added a test to ensure that #77 really is fixed. Feel free to review or merge.

PR #90, opened 2022-01-23
Issue hunt2

### LINKED ISSUE #81: two problems with joining new channel and kicking self

ThePhilgrim:
Fixes #81

### DIFF
--- a/commands.py
+++ b/commands.py
@@ -208,10 +208,10 @@ def handle_kick(state: mantatail.ServerState, user: mantatail.UserConnection, ar
         return
 
     if len(args) == 2:
-        message = f"KICK {channel.name} {target_usr.nick} :{target_usr.nick}\r\n"
+        message = f"KICK {channel.name} {target_usr.nick} :{target_usr.nick}"
     elif len(args) >= 3:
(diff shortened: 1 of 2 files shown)


commands.py
```diff
@@ -199,7 +199,7 @@ def handle_kick(state: mantatail.ServerState, user: mantatail.UserConnection, ar
         error_no_such_nick_channel(user, args[1])
         return
 
-    if user not in channel.operators:
+    if not (channel.is_founder(user) or user in channel.operators):
```
Akuli, line 202: I think the founder doesn't necessarily have to be an operator, and in that case, won't have the permissions that an operator has. It might be good to check this on libera or ircdocs.horse.

Deleting the channel when the last user leaves should be enough to fix #81.

Akuli: We decided to remove founder feature later

PR #92, opened 2022-01-23
Issue hunt3

### LINKED ISSUE #78: kicking doesn't drop operator privileges
Consider the following situation:
1. `x` creates a channel `#foo` and becomes the founder
2. `y` joins `#foo`
3. `x` gives channel operator status to `y` on `#foo`
4. `x` kicks `y` from `#foo`
5. `y` joins back to `#foo`

This situation can happen in practice, for example, if `y` abuses their operator permissions in some way, and `x` decides it was a bad idea to make `y` an operator. So the reasonable behavior would be that when `y` joins back, `y` is no longer an operator.

But that's not what happens: if an operator is kicked from a channel and they join back, they are still an operator.

ThePhilgrim:
Fixes #78

### DIFF
--- a/commands.py
+++ b/commands.py
@@ -215,8 +215,7 @@ def handle_kick(state: mantatail.ServerState, user: mantatail.UserConnection, ar
 
     channel.kick_user(user, target_usr, message)
 
-    if target_usr in channel.operators:
-        channel.operators.discard(target_usr)
+    channel.operators.discard(target_usr)
 
     if len(channel.users) == 0:
         state.delete_channel(channel.name)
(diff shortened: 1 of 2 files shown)


commands.py
```diff
@@ -215,6 +215,9 @@ def handle_kick(state: mantatail.ServerState, user: mantatail.UserConnection, ar
 
     channel.kick_user(user, target_usr, message)
 
+    if target_usr in channel.operators:
+        channel.operators.discard(target_usr)
```
Akuli, line 219: `discard` does nothing if the user is not in the set. This is how `discard` and `remove` differ.
```suggestion
    channel.operators.discard(target_usr)
```

PR #93, opened 2022-01-23
Issue hunt4

### LINKED ISSUE #76: "MODE +o #foo y" fails silently

### LINKED ISSUE #88: RPL_CHANNELMODEIS is broken

ThePhilgrim:
Fixes #76
Fixes #88

### DIFF
--- a/commands.py
+++ b/commands.py
@@ -351,7 +351,10 @@ def process_channel_modes(state: mantatail.ServerState, user: mantatail.UserConn
         return
 
     if len(args) == 1:
-        message = f'{irc_responses.RPL_CHANNELMODEIS} {user.nick} {channel.name} {" ".join(channel.modes)}'
+        if channel.modes:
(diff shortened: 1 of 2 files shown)


Akuli:
Please add a test where the server sends RPL_CHANNELMODEIS.

Akuli (approved):

PR #97, opened 2022-01-23
Fuzzer update

ThePhilgrim:
By keeping `sock.connect()` in a try/except, the user can be notified straight in the terminal that they forgot to start mantatail. I believe this is more effective that an inline comment, that actually requires you to open the file.

This would have helped me when I tried to run the fuzzer the first time without realizing how to use it, and without reading the actual script.

### DIFF
--- a/tests/fuzzer/fuzzer.py
+++ b/tests/fuzzer/fuzzer.py
@@ -8,6 +8,7 @@ Usage:
 
 import socket
 import random
+import sys
 
 
 words = [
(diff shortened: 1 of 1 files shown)


tests/fuzzer/fuzzer.py
```diff
@@ -40,7 +41,11 @@
         command += " ".join(chosen_words) + "\n"
 
     sock = socket.socket()
-    sock.connect(("localhost", 6667))  # If this fails you forgot to start mantatail
+    try:
+        sock.connect(("localhost", 6667))
+    except:
```
Akuli, line 46: Be specific about what error you're catching. If it's a KeyboardInterrupt or a bad file descriptor error, we'd rather just show the full error to user.

```suggestion
    except ConnectionRefusedError:
```

Akuli:
> that actually requires you to open the file.

It does not. The comment shows up in the error message.

Akuli:
But that said, this PR is still a big improvement :)

ThePhilgrim:
> > that actually requires you to open the file.
> 
> It does not. The comment shows up in the error message.

Okay. It does, however, get less clear since the comment is buried in the rather long error message.

ThePhilgrim:
I guess it could be even clearer if a successful run would do `print("Fuzzer running. Please switch to Mantatail terminal to see eventual issues`.

Akuli:
Detecting a successful run is a bit difficult. We can't put the print in the loop, because then it will get printed many times. If we print before the loop, it will get printed even if the server isn't running. If we put it after the loop, the print won't run at all.

Maybe just put it before the loop, but instead of "Fuzzer running" print "Fuzzer starting".

ThePhilgrim:
I guess we could put the while loop in a function, and having the print before the while loop, in the function?

PR #98, opened 2022-01-23
fuzzer start print

Akuli:
I think this makes the fuzzer as user friendly as it will ever have to be.

### DIFF
--- a/tests/fuzzer/fuzzer.py
+++ b/tests/fuzzer/fuzzer.py
@@ -33,6 +33,8 @@ words = [
     "",
 ]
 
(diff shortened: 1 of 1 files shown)


ThePhilgrim:
@Akuli As I said in the previous PR: Wouldn't it make sense to put the while loop in a function, and have the print in the beginning of the while loop? That way we know what the server is definitely running when the print runs

Akuli:
A function won't help: if the print is in the beginning of the loop, it will run every time, regardless of whether it's in a function or not. I also don't think we should make the script too perfect, just easy enough to use without reading the instructions.

PR #99, opened 2022-01-23
Move KeyError checks to state methods

### LINKED ISSUE #72: KeyErrors are raised in `find_user()` & `find_channel()`
Right now, calling these methods always have to be done in a `try/except`.

Ex:
```
try:
    find_user(Alice)
except KeyError:
    # handle error here
```

It would be better to have the methods return `None` if there is no user/channel found instead.

ThePhilgrim:
Fixes #72

### DIFF
--- a/commands.py
+++ b/commands.py
@@ -132,6 +132,17 @@ def handle_mode(state: mantatail.ServerState, user: mantatail.UserConnection, ar
     if args[0].startswith("#"):
         process_channel_modes(state, user, args)
     else:
+        try:
+            target_usr = state.find_user(args[0])
+            if user != target_usr:
+                # TODO: The actual IRC error for this should be "502 Can't change mode for other users"
+                # This will be implemented when MODE becomes more widely supported.
+                # Currently not sure which modes 502 applies to.
+                error_no_such_channel(user, args[0])
+                return
+        except KeyError:
+            error_no_such_channel(user, args[0])
+            return
         process_user_modes()
 
 
@@ -333,15 +344,6 @@ def process_channel_modes(state: mantatail.ServerState, user: mantatail.UserConn
 
     Finally sends a message to all users on the channel, notifying them about the new channel mode.
     """
-    if args[1][0] not in ["+", "-"]:
-        error_unknown_mode(user, args[1][0])
-        return
-    supported_modes = ["o"]
-    for mode in args[1][1:]:
-        if mode not in supported_modes:
-            error_unknown_mode(user, mode)
-            return
-
     try:
         channel = state.find_channel(args[0])
     except KeyError:
@@ -349,11 +351,23 @@ def process_channel_modes(state: mantatail.ServerState, user: mantatail.UserConn
         return
 
     if len(args) == 1:
-        message = f'{irc_responses.RPL_CHANNELMODEIS} {user.nick} {channel.name} {" ".join(channel.modes)}'
+        if channel.modes:
+            message = f'{irc_responses.RPL_CHANNELMODEIS} {user.nick} {channel.name} {" ".join(channel.modes)}'
+        else:
+            message = f"{irc_responses.RPL_CHANNELMODEIS} {user.nick} {channel.name}"
         user.send_que.put((message, "mantatail"))
     elif len(args) == 2:
         error_not_enough_params(user, "MODE")
     else:
+        if args[1][0] not in ["+", "-"]:
+            error_unknown_mode(user, args[1][0])
+            return
+        supported_modes = ["o"]
+        for mode in args[1][1:]:
+            if mode not in supported_modes:
+                error_unknown_mode(user, mode)
+                return
+
         mode_command, flags = args[1][0], args[1][1:]
         try:
             target_usr = state.find_user(args[2])
--- a/commands.py
+++ b/commands.py
@@ -53,26 +53,27 @@ def handle_join(state: mantatail.ServerState, user: mantatail.UserConnection, ar
 
         channel = state.find_channel(channel_name)
 
-        if channel:
-            if user not in channel.users:
-                channel_users_str = ""
-                for usr in channel.users:
-                    channel_users_str += f" {usr.get_nick_with_prefix(channel)}"
+        assert channel
 
-                channel.users.add(user)
+        if user not in channel.users:
+            channel_users_str = ""
+            for usr in channel.users:
+                channel_users_str += f" {usr.get_nick_with_prefix(channel)}"
 
-                for usr in channel.users:
-                    message = f"JOIN {channel_name}"
-                    usr.send_que.put((message, user.get_user_mask()))
+            channel.users.add(user)
 
-                # TODO: Implement topic functionality for existing channels & MODE for new ones
+            for usr in channel.users:
+                message = f"JOIN {channel_name}"
+                usr.send_que.put((message, user.get_user_mask()))
 
-                message = f"353 {user.nick} = {channel_name} :{user.get_nick_with_prefix(channel)}{channel_users_str}"
+            # TODO: Implement topic functionality for existing channels & MODE for new ones
 
-                user.send_que.put((message, "mantatail"))
+            message = f"353 {user.nick} = {channel_name} :{user.get_nick_with_prefix(channel)}{channel_users_str}"
 
-                message = f"366 {user.nick} {channel_name} :End of /NAMES list."
-                user.send_que.put((message, "mantatail"))
+            user.send_que.put((message, "mantatail"))
+
+            message = f"366 {user.nick} {channel_name} :End of /NAMES list."
+            user.send_que.put((message, "mantatail"))
 
         # TODO:
         #   * Send topic (332)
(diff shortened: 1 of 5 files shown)


commands.py
```diff
@@ -180,7 +180,9 @@ def handle_nick(state: mantatail.ServerState, user: mantatail.UserConnection, ar
             if user.user_message:
                 user.send_que.put((message, user.get_user_mask()))
 
+            # Not using state.delete_user() as that will delete the user from all channels as well.
             del state.connected_users[user.nick.lower()]
```
ThePhilgrim, line 184: This is a problematic line. It creates a `KeyError` over and over when running the fuzzer. I mean, I guess I could just make a try/except KeyError. But I have the feeling that's not solving the issue at the root.

The problem is I can't see what command the Fuzzer actually sent to mantatail. So it's hard to debug.

Can we make the fuzzer print every command it runs, so that it might be possible to retrace the issue?

ThePhilgrim:
As we see here, I get a lot of `KeyError`s from that line. I also get `IndexError`s still from `handle_mode`

```
 process_channel_modes
    del state.connected_users[user.nick.lower()]
KeyError: 'nick'
IndexError: string index out of range
Exception in thread Thread-1297:
  File "/usr/local/Cellar/python@3.9/3.9.7_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/threading.py", line 973, in _bootstrap_inner
    self.run()
  File "/usr/local/Cellar/python@3.9/3.9.7_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/threading.py", line 910, in run
  File "/Users/Writing/pthn/mantatail/mantatail.py", line 199, in recv_loop
    call_handler_function(state, user, args)
  File "/Users/Writing/pthn/mantatail/commands.py", line 133, in handle_mode
    process_channel_modes(state, user, args)
  File "/Users/Writing/pthn/mantatail/commands.py", line 361, in process_channel_modes
IndexError: string index out of range
Exception in thread Thread-1307:
  File "/usr/local/Cellar/python@3.9/3.9.7_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/threading.py", line 973, in _bootstrap_inner
    self.run()
  File "/usr/local/Cellar/python@3.9/3.9.7_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/threading.py", line 910, in run
^CTraceback (most recent call last):
  File "/usr/local/Cellar/python@3.9/3.9.7_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/runpy.py", line 197, in _run_module_as_main
  File "/Users/Writing/pthn/mantatail/mantatail.py", line 199, in recv_loop
    return _run_code(code, main_globals, None,
  File "/usr/local/Cellar/python@3.9/3.9.7_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/runpy.py", line 87, in _run_code
    call_handler_function(state, user, args)
  File "/Users/Writing/pthn/mantatail/commands.py", line 184, in handle_nick
    exec(code, run_globals)
  File "/Users/Writing/pthn/mantatail/mantatail.py", line 412, in <module>
    del state.connected_users[user.nick.lower()]
KeyError: 'pong'
```

commands.py
Akuli: You can add something like `print(command, "----------------------")` to the fuzzer's loop. It will print a LOT of stuff, though, as everything it sends consists of 1000 lines. Here are a couple ideas that might help:
- Make the inputs consist of e.g. 50 lines instead of 1000 lines
- Add `input("Press enter to continue...")` to the end of the loop, so that you can ask it to send the next commands only if you see an error.

commands.py
```diff
@@ -51,27 +51,28 @@ def handle_join(state: mantatail.ServerState, user: mantatail.UserConnection, ar
...
             state.channels[lower_channel_name] = mantatail.Channel(channel_name, user)
 
-        channel = state.channels[lower_channel_name]
+        channel = state.find_channel(channel_name)
 
-        if user not in channel.users:
-            channel_users_str = ""
-            for usr in channel.users:
-                channel_users_str += f" {usr.get_nick_with_prefix(channel)}"
+        if channel:
```
Akuli, line 56: - How can the channel be None? If the channel doesn't exist, it is created above, so when you call `find_channel()`, it should now exist.
- `if not channel: return` would make the code more readable, because then the rest of the code doesn't have to be indented.

commands.py
```diff
@@ -94,24 +95,23 @@ def handle_part(state: mantatail.ServerState, user: mantatail.UserConnection, ar
...
-    except KeyError:
-        error_no_such_channel(user, channel_name)
-        return
+    channel = state.find_channel(channel_name)
 
-    if user not in channel.users:
-        error_not_on_channel(user, channel_name)
+    if not channel:
+        error_no_such_channel(user, channel_name)
     else:
```
Akuli, line 102: This too would be simpler with a `return` statement.

mantatail.py
```diff
@@ -48,15 +60,22 @@ def delete_user(self, nick: str) -> None:
         Note: This does not actually disconnect the user from the server.
         To disconnect the user, a tuple (None, None) must be put in their send queue.
         """
-        user = self.connected_users[nick.lower()]
-        for channel in self.channels.values():
-            if user in channel.users:
-                channel.users.discard(user)
-        del self.connected_users[nick.lower()]
+        user = self.find_user(nick)
+        if user:
```
Akuli, line 64: When is `delete_user()` called with a non-existing user? I think that should be an error, so `assert user is not None` is more appropriate than `if user:`.

mantatail.py
```diff
@@ -48,15 +60,22 @@ def delete_user(self, nick: str) -> None:
...
+                if user in channel.users:
+                    channel.users.discard(user)
+            del self.connected_users[nick.lower()]
 
     def delete_channel(self, channel_name: str) -> None:
-        """Removes a channel from server."""
-        del self.channels[channel_name.lower()]
+        """
+        Removes a channel from server.
+        Silently ignores a non-existing channel.
```
Akuli, line 73: Why is silently ignoring a good thing?

Akuli (commented):
As you said, adding `if` statements near erroring places is generally not a good way to fix errors. But the same applies to other lines than the one you pointed out: they add new if statements that just silently do nothing if a user or channel is not found. That's not a good thing, because it can make debugging more difficult.

In general, if something is wrong, you want to get an error as early as possible, because errors happening earlier are closer to the root cause of the problem. If your mistake messes up something else that then eventually errors something seemingly unrelated, that's not exactly very easy to debug. I know this too well as a C programmer :)

commands.py
ThePhilgrim: @Akuli I tried to add some prints to the loop 2 days ago. But I noticed that it stopped giving me the errors when I did that. Do you have an idea why this might be?

mantatail.py
ThePhilgrim: It's not necessarily a good or bad thing. I didn't know how else to treat it.

The user never deliberately deletes a channel. If they `PART` as the last user on the channel, it is deleted.
Therefore, there should never be a `KeyError`.. which I realize now, is a good reason to instead `raise` an error if it would happen anyways.

mantatail.py
```diff
@@ -48,15 +60,24 @@ def delete_user(self, nick: str) -> None:
...
+        """
+        try:
+            del self.channels[channel_name.lower()]
+        except KeyError as err:
+            print(err)
+            print("Tried to delete a channel that does not exist.")
```
Akuli, line 80: Just let Python handle printing an error message. It does better job than you did. The problem is that `print(err)` shows a very short error message, such as `'#mantatest'`, whereas Python by default would show multiple lines ending with `KeyError: '#mantatest'`.

For example:

```
>>> channels = {}
>>> try:
...     del channels["#mantatest"]
... except KeyError as err:
...     print(err)
... 
'#mantatest'
```

```suggestion
        del self.channels[channel_name.lower()]
```

commands.py
ThePhilgrim: Hm, it seems to give me errors now even if I got some prints!

However, could you explain to me why the amount of commands it sends increase instead of sending one command at a time?

After changing to `for line_number in range(5):`, I get this output

Skärmavbild 2022-01-25 kl  15 32 54

As you see, there are more and more commands before the `------------`.

mantatail.py
ThePhilgrim: However, wouldn't the default Python error break/interrupt the server, whereas in an `except, the server keeps running? That's how I thought it worked.

Akuli: If you really want the server to continue running, even though its internal state seems to be broken in an unexpected way, you can use `traceback.print_exc()` to print a full error message. But that should be done in a separate PR.

ThePhilgrim: Okay, would you say it makes more sense to just let the server break if this error occurs?

ThePhilgrim:
@Akuli I added a print in `handle_nick()` that shows what is received when it errors. It doesn't make a lot of sense to me though.

Skärmavbild 2022-01-25 kl  15 49 02

As you can see, it errors with `KeyError: 'privmsg'`. However, the last command sent to `handle_nick()` contained only `MODE`, not `PRIVMSG`. 

Any thoughts on this?

mantatail.py
Akuli: I would much rather have the server interrupt itself than e.g. let non-operators ban other users, or do something else that the server might end up doing if the state is messed up.

Akuli:
After staring at this for a while, I realized that the print is before setting the nick of the user. So the old nick of the user is `PRIVMSG`, and we are trying to changing its nick to `MODE`.

At this point, I would make a text file that contains the commands the fuzzer sent, so that you can send the same exact commands again easily. If you have the commands saved in a file `commands.txt`, you can redirect it as input to netcat like this:

```
nc localhost 6667 < commands.txt
```

This tells the shell to copy/paste the contents of commands.txt into netcat. Also, send me the exact commands, so that I don't have to type them manually from your screenshot and guess what was before the visible part :)

ThePhilgrim:
@Akuli I tried to do as you said. However, when running the file via netcat, there are no errors at all. It just says "`:QUIT :Quit: (Remote host closed the connection)`". I tried to remove all instances of `QUIT`, but it changes nothing.

Here is the whole set of commands run by the fuzzer:
https://dpaste.com/GTXFPHXC6

mantatail.py
```diff
@@ -48,14 +60,19 @@ def delete_user(self, nick: str) -> None:
...
         for channel in self.channels.values():
             if user in channel.users:
                 channel.users.discard(user)
         del self.connected_users[nick.lower()]
 
     def delete_channel(self, channel_name: str) -> None:
-        """Removes a channel from server."""
+        """
+        Removes a channel from server.
+        Silently ignores a non-existing channel.
```
Akuli, line 74:
```suggestion
```

PR #101, opened 2022-01-26
Change logic for sending message to channel users

ThePhilgrim:
I think this improves the way messages are sent to channels. That way I don't have to repeat the same for loop in every single `handle_` function.

### DIFF
--- a/commands.py
+++ b/commands.py
@@ -273,7 +273,7 @@ def handle_privmsg(state: mantatail.ServerState, user: mantatail.UserConnection,
         error_not_on_channel(user, receiver)
     else:
         privmsg_message = f"PRIVMSG {receiver} :{privmsg}"
-        channel.queue_message_to_chan_users(privmsg_message, user, False)
+        channel.queue_message_to_chan_users(privmsg_message, user, send_to_self=False)
 
 
 def handle_pong(state: mantatail.ServerState, user: mantatail.UserConnection, args: List[str]) -> None:
--- a/commands.py
+++ b/commands.py
@@ -62,9 +62,8 @@ def handle_join(state: mantatail.ServerState, user: mantatail.UserConnection, ar
 
             channel.users.add(user)
 
-            for usr in channel.users:
-                message = f"JOIN {channel_name}"
-                usr.send_que.put((message, user.get_user_mask()))
+            join_msg = f"JOIN {channel_name}"
+            channel.queue_message_to_chan_users(join_msg, user)
 
             # TODO: Implement topic functionality for existing channels & MODE for new ones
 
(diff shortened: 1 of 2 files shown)


commands.py
```diff
@@ -273,10 +272,8 @@ def handle_privmsg(state: mantatail.ServerState, user: mantatail.UserConnection,
     if user not in channel.users:
         error_not_on_channel(user, receiver)
     else:
-        for usr in channel.users:
-            if usr.nick != user.nick:
-                message = f"PRIVMSG {receiver} :{privmsg}"
-                usr.send_que.put((message, user.get_user_mask()))
+        privmsg_message = f"PRIVMSG {receiver} :{privmsg}"
+        channel.queue_message_to_chan_users(privmsg_message, user, False)
```
Akuli, line 276: If you use a keyword argument, the reader doesn't have to guess what `False` does:

```suggestion
        channel.queue_message_to_chan_users(privmsg_message, user, send_to_self=False)
```

mantatail.py
```diff
@@ -375,15 +375,23 @@ def is_founder(self, user: UserConnection) -> bool:
         """Checks if the user is the channel founder."""
         return user.user_name == self.founder
 
-    def kick_user(self, kicker: UserConnection, user_to_kick: UserConnection, message: str) -> None:
+    def kick_user(self, user_to_kick: UserConnection) -> None:
+        """Kicks user from the channel."""
+        self.users.discard(user_to_kick)
```
Akuli, line 380: This method has now become so simple that it should just be removed IMO.

mantatail.py
```diff
@@ -375,15 +375,23 @@ def is_founder(self, user: UserConnection) -> bool:
...
         """
         for usr in self.users:
-            usr.send_que.put((message, kicker.get_user_mask()))
-
-        self.users.discard(user_to_kick)
+            if not send_to_self:
+                if usr != sender:
+                    usr.send_que.put((message, sender.get_user_mask()))
+            else:
+                usr.send_que.put((message, sender.get_user_mask()))
```
Akuli, line 394: Maybe it's simpler if you think about it like this: we want to send if it's a different user, or it's the same user but we are allowed to send to self.

```suggestion
            if usr != sender or send_to_self:
                usr.send_que.put((message, sender.get_user_mask()))
```

Akuli (commented):
I like the idea! A few nits below.

commands.py
ThePhilgrim: Ah, nice! I didn't know that you could use the keyword in the argument too. I thought it was only in the parameters.

mantatail.py
ThePhilgrim: Very clever! I like it

PR #102, opened 2022-01-26
Channel topic

ThePhilgrim:
This PR implements the logic for channel topics.

- Test has been added.
- Topic keyword has been added to fuzzer.
- I get no errors when running the fuzzer between 30sec to 1 min


Akuli (approved):
Works in mantaray

PR #103, opened 2022-01-26
Fuzzer improvements

Akuli:
- When Mantatail crashes, print the commands that were sent to it
- Run Mantatail automatically, no longer two separate terminals needed
- Add `:` to the allowed words

### DIFF
--- a/tests/fuzzer/fuzzer.py
+++ b/tests/fuzzer/fuzzer.py
@@ -139,8 +139,8 @@ def fuzzing_loop():
 mantatail_process = subprocess.Popen(
     ["python3", "-u", "mantatail.py"], cwd="../..", stdout=subprocess.PIPE, stderr=subprocess.STDOUT
 )
-print(mantatail_process.stdout.readline())  # Wait for mantatail to start
 try:
+    print(mantatail_process.stdout.readline())  # Wait for mantatail to start
     threading.Thread(target=output_reading_thread).start()
     fuzzing_loop()
 finally:
--- a/tests/fuzzer/fuzzer.py
+++ b/tests/fuzzer/fuzzer.py
@@ -114,13 +114,8 @@ def fuzzing_loop():
             commands += " ".join(chosen_words) + "\n"
 
         sock = socket.socket()
-        try:
-            sock.connect(("localhost", 6667))
-        except ConnectionRefusedError:
-            sys.exit("Connection Refused: Start Mantatail in a separate terminal before running fuzzer.py")
-
+        sock.connect(("localhost", 6667))
         recent_commands.append((sock.getsockname(), commands))
-
         sock.sendall(commands.encode())
 
         try:
--- a/tests/fuzzer/fuzzer.py
+++ b/tests/fuzzer/fuzzer.py
@@ -15,7 +15,6 @@ import collections
 import random
 import socket
 import subprocess
-import sys
 import threading
 
 
--- a/tests/fuzzer/fuzzer.py
+++ b/tests/fuzzer/fuzzer.py
@@ -12,10 +12,16 @@ To fuzz:
 """
 
 import collections
+import os
 import random
 import socket
 import subprocess
 import threading
+from pathlib import Path
+
+# Change to the directory where this script is, so it doesn't matter whether
+# you run "python3 fuzzer.py" or "python3 tests/fuzzer/fuzzer.py"
+os.chdir(Path(__file__).parent)
 
 
 words = [
--- a/tests/fuzzer/fuzzer.py
+++ b/tests/fuzzer/fuzzer.py
@@ -83,7 +83,7 @@ def print_commands(source_string):
     recent_commands_copy = list(recent_commands)
 
     for address_tuple, commands in recent_commands_copy:
-        if str(address_tuple) == source_string:  # lol
+        if str(address_tuple) == source_string:
             print(commands)
 
 
--- a/tests/fuzzer/fuzzer.py
+++ b/tests/fuzzer/fuzzer.py
@@ -1,14 +1,22 @@
 """Simple fuzzer for MantaTail. See https://en.wikipedia.org/wiki/Fuzzing
 
-Usage:
-- Start MantaTail: python3 -m mantatail
-- Run fuzzer in another terminal: python3 fuzzer.py
-- Go back to MantaTail terminal, and see if you get errors.
+To fuzz:
+
+ 1. Make sure you don't have any unnecessary prints in mantatail. They will
+    cause the fuzzer to think that mantatail crashed; it doesn't try to
+    distinguish prints from error messages.
+
+ 2. Make sure that Mantatail isn't running.
+
+ 3. Run this script.
 """
 
-import socket
+import collections
 import random
+import socket
+import subprocess
 import sys
+import threading
 
 
 words = [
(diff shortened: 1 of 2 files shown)


tests/fuzzer/fuzzer.py
```diff
@@ -1,14 +1,21 @@
...
-- Run fuzzer in another terminal: python3 fuzzer.py
-- Go back to MantaTail terminal, and see if you get errors.
+To fuzz:
+
+ 1. Make sure you don't have any unnecessary prints in mantatail. They will
+    cause the fuzzer to think that mantatail crashed; it doesn't try to
+    distinguish prints from error messages.
+
+ 2. Make sure that Mantatail isn't running.
+
```
ThePhilgrim, line 10: If I `cd` to `mantatail` folder and run `python3 tests/fuzzer/fuzzer.py`, the connection is refused because the path for running `mantatail.py` gets messed up.

To make the fuzzer work, I need to `cd` to `mantatail/tests/fuzzer` and run the script. Please add this to the instructions.

tests/fuzzer/fuzzer.py
```diff
@@ -32,28 +39,103 @@
...
+#    IndexError: string index out of range
+#    Got connection from ('127.0.0.1', 33232)
+#    Got connection from ('127.0.0.1', 33234)
+#    Got connection from ('127.0.0.1', 33236)
+#    Got connection from ('127.0.0.1', 33238)
+#
+# To figure out what commands caused the crash, i.e. what was sent from address
+# tuple ('127.0.0.1', 33230), we keep the most recent address tuples and
+# commands here.
+recent_commands = collections.deque(maxlen=100)
```
ThePhilgrim, line 72: Do we really need 100 commands in the stack? When the server errors, we're most likely only interested in the very last commands. At that point, the earlier commands are irrelevant, and might even cause confusion. For example, when I run the fuzzer now, I get the following error:

```
PART
USER -o PING JOIN
PART TOPIC -g KICK MODE
: NICK PONG : PRIVMSG
NICK
QUIT -o #foo USER :
PONG #bar
TOPIC PRIVMSG PART PART QUIT
+o PRIVMSG PRIVMSG
-o TOPIC -g -g PING
#bar JOIN #foo -xyz :
-o USER KICK -xyz -g
-xyz
-xyz +xyz -o
 #foo
MODE QUIT TOPIC
JOIN PRIVMSG KICK PRIVMSG

#foo
#foo MODE PONG PART :
KICK +o -o MODE
JOIN TOPIC -g PING
+xyz
+xyz
-g
+o
QUIT QUIT :
-xyz : QUIT  #bar
#bar
PONG #bar : QUIT PONG
PRIVMSG #bar
NICK
-o
-xyz
: :
PART : JOIN NICK
+o
PRIVMSG JOIN USER
PONG

*** Errors: ***
    call_handler_function(state, user, args)
  File "/Users/Writing/pthn/mantatail/commands.py", line 133, in handle_mode
    process_channel_modes(state, user, args)
  File "/Users/Writing/pthn/mantatail/commands.py", line 396, in process_channel_modes
IndexError: string index out of range

-------- CRASH END --------
```

The commands continue up very far. It makes me wonder what actually caused the error.

tests/fuzzer/fuzzer.py
```diff
@@ -32,28 +39,103 @@
...
+    recent_commands_copy = list(recent_commands)
 
-while True:
-    command = ""
-    for line_number in range(10000):
-        words_per_line = random.randint(1, 5)
-        chosen_words = [random.choice(words) for word_number in range(words_per_line)]
-        command += " ".join(chosen_words) + "\n"
+    for address_tuple, commands in recent_commands_copy:
+        if str(address_tuple) == source_string:  # lol
```
ThePhilgrim, line 80: Why the `# lol`?

tests/fuzzer/fuzzer.py
```diff
@@ -32,28 +39,103 @@
...
+            print(commands)
 
-    sock = socket.socket()
-    try:
+
+def output_reading_thread():
+    source = None
+    output_lines = []
+
+    for line in mantatail_process.stdout:
```
ThePhilgrim, line 88: I think I understand 90% of how the fuzzer works. The things I don't understand are details.

One thing that I am wondering is:
The `output_reading_thread` starts before the `fuzzing_loop()` starts. The output reading does not happen in a while loop. This means that when `for line in mantatail_process.stdout` runs, it should be empty. Then, how does it run again? There is not a separate thread for each connection, so in my mind, `output_reading_thread` only runs once.

tests/fuzzer/fuzzer.py
```diff
@@ -32,28 +39,103 @@
...
+    while True:
+        commands = ""
+        for line_number in range(500):
+            words_per_line = random.randint(1, 5)
+            chosen_words = [random.choice(words) for word_number in range(words_per_line)]
+            commands += " ".join(chosen_words) + "\n"
+
+        sock = socket.socket()
         sock.connect(("localhost", 6667))
-    except ConnectionRefusedError:
```
ThePhilgrim, line 49: I would add this `try/except` back with `sys.exit("Connection Refused: Please make sure you are located in the tests/fuzzer directory")` or something similar (see comment above for reason).

ThePhilgrim (changes requested):

tests/fuzzer/fuzzer.py
Akuli: There is actually 500 lines of commands, not 100 lines of commands. When the fuzzing loop has sent the 500 lines of commands, it puts **one** element to the deque. So the deque maxlen doesn't have anything to do with how many commands are sent; they all are printed from a single deque element anyway.

You can try to change 500 below to a smaller number, but then the fuzzer no longer found crashes, at least to me. I have an idea that can help with this, but I'd rather change it in a separat ePR.

The deque is needed because the fuzzer might send several new sets of 500 commands before it sees the errors coming from mantatail. It just has to be big enough that the sent commands are still there when an error is detected. I tried `maxlen=10`, and it didn't work every time.

tests/fuzzer/fuzzer.py
Akuli: It is relying on how tuples look when converted to strings. Not great, but also not too bad for testing purposes.

tests/fuzzer/fuzzer.py
Akuli: The for loop reads output line by line as long as the process is running. When mantatail doesn't print anything, the for loop is waiting and doesn't do anything.

As usual, this:

```
for line in file:
    do stuff with line
```

...is really same as this:

```
while True:
    line = file.readline()  # This will wait until a new line is printed
    if not line:
        break
    do stuff with line
```

tests/fuzzer/fuzzer.py
Akuli: I changed the fuzzer to `cd` automatically.

tests/fuzzer/fuzzer.py
ThePhilgrim: Hm, but 

```
for line in file:
    do stuff with line
```

wouldn't do anything if `file` was empty. In my mind, this would just be like doing

```
for x in range(0):
   print("do stuff here")
```

Akuli: We had a lengthy discussion about how "files" work, and how reading/writing "files" can behave very differently from regular files.

ThePhilgrim (approved):

PR #104, opened 2022-01-26
Implement logic for quit messages/reasons

ThePhilgrim:
Since the `message` queued in a users `send_que` is never `None`, I decided that the second `None` in the tuple was not necessarily needed. Therefore I saw it as the best option to send a quit message/reason.

### DIFF
--- a/mantatail.py
+++ b/mantatail.py
@@ -142,7 +142,7 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
     """
 
     user = UserConnection(state, user_host, user_socket)
-
+    disconnect_reason = ""
     try:
         while True:
             request = b""
@@ -150,7 +150,8 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
                 user.start_ping_timer()
                 try:
                     request_chunk = user_socket.recv(4096)
-                except OSError:
+                except OSError as err:
+                    disconnect_reason = err.strerror
                     return  # go to "finally:"
                 finally:
                     user.ping_timer.cancel()
@@ -158,6 +159,7 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
                 if request_chunk:
                     request += request_chunk
                 else:
+                    disconnect_reason = "Remote host closed the connection"
                     return  # go to "finally:"
 
             decoded_message = request.decode("latin-1")
@@ -179,6 +181,7 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
                         commands.handle_pong(state, user, args)
                     else:
                         if command_lower == "quit":
+                            disconnect_reason = "Remote host closed the connection"
                             return
                         else:
                             commands.error_not_registered(user)
@@ -198,7 +201,6 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
                             if command_lower == "quit":
                                 return
     finally:
-        disconnect_reason = "Remote host closed the connection"
         user.send_que.put((None, disconnect_reason))
 
 
@@ -280,8 +282,8 @@ class UserConnection:
             else:
                 try:
                     self.send_string_to_client(message, prefix)
-                except:
-                    disconnect_reason = "Remote host closed the connection"
+                except OSError as err:
+                    disconnect_reason = err.strerror
                     self.send_que.put((None, disconnect_reason))
 
     def queue_quit_message_for_other_users(self, quit_message: str) -> None:
--- a/mantatail.py
+++ b/mantatail.py
@@ -198,7 +198,7 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
                             if command_lower == "quit":
                                 return
     finally:
-        disconnect_reason = "(Remote host closed the connection)"
+        disconnect_reason = "Remote host closed the connection"
         user.send_que.put((None, disconnect_reason))
 
 
@@ -232,7 +232,7 @@ class UserConnection:
         self.nick = "*"
         self.user_message: Optional[List[str]] = None  # Ex. AliceUsr 0 * Alice
         self.user_name: Optional[str] = None  # Ex. AliceUsr
-        self.send_que: queue.Queue[Tuple[str, str] | Tuple[None, Optional[str]]] = queue.Queue()
+        self.send_que: queue.Queue[Tuple[str, str] | Tuple[None, str]] = queue.Queue()
         self.que_thread = threading.Thread(target=self.send_queue_thread)
         self.que_thread.start()
         self.pong_received = False
@@ -260,7 +260,7 @@ class UserConnection:
 
             if message is None:
                 disconnect_reason = prefix
-                quit_message = f"QUIT :Quit: {disconnect_reason}"
+                quit_message = f"QUIT :Quit: ({disconnect_reason})"
                 with self.state.lock:
                     self.queue_quit_message_for_other_users(quit_message)
                     if self.nick != "*":
@@ -281,7 +281,7 @@ class UserConnection:
                 try:
                     self.send_string_to_client(message, prefix)
                 except:
-                    disconnect_reason = "(Remote host closed the connection)"
+                    disconnect_reason = "Remote host closed the connection"
                     self.send_que.put((None, disconnect_reason))
 
     def queue_quit_message_for_other_users(self, quit_message: str) -> None:
@@ -348,7 +348,7 @@ class UserConnection:
         If no PONG response has been received, the server closes the connection to the client.
         """
         if not self.pong_received:
-            disconnect_reason = "(Ping timeout...)"
+            disconnect_reason = "Ping timeout..."
             self.send_que.put((None, disconnect_reason))
         else:
             self.pong_received = False
--- a/mantatail.py
+++ b/mantatail.py
@@ -59,7 +59,7 @@ class ServerState:
         thereafter removes user from connected users.
 
         Note: This does not actually disconnect the user from the server.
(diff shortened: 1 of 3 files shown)


mantatail.py
```diff
@@ -198,7 +198,8 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
                             if command_lower == "quit":
                                 return
     finally:
-        user.send_que.put((None, None))
+        disconnect_reason = "(Remote host closed the connection)"
```
Akuli, line 201: I believe "Remote host closed the connection" means that the user closed their end of the socket connection, which is not necessarily what happened here. When receiving fails, the QUIT reason should IMO include the error message from `recv()`, e.g. "Connection reset by peer" or "Broken pipe".

Also, I don't like how you have to make sure to include `"("` and `")"` when you put a disconnect reason to the queue. Can we instead add it in `send_queue_thread`?

mantatail.py
```diff
@@ -231,7 +232,7 @@ def __init__(self, state: ServerState, host: str, socket: socket.socket):
         self.nick = "*"
         self.user_message: Optional[List[str]] = None  # Ex. AliceUsr 0 * Alice
         self.user_name: Optional[str] = None  # Ex. AliceUsr
-        self.send_que: queue.Queue[Tuple[str, str] | Tuple[None, None]] = queue.Queue()
+        self.send_que: queue.Queue[Tuple[str, str] | Tuple[None, Optional[str]]] = queue.Queue()
```
Akuli, line 235: Can the second element of the tuple ever be None?

mantatail.py
```diff
@@ -280,21 +281,18 @@ def send_queue_thread(self) -> None:
                 try:
                     self.send_string_to_client(message, prefix)
                 except:
-                    self.send_que.put((None, None))
+                    disconnect_reason = "(Remote host closed the connection)"
```
Akuli, line 284: Same here, this should include the error message.

Also, a bare `except:` is bad because it also catches e.g. `KeyboardInterrupt` and `AttributeError`, and that can make debugging more difficult. Always* specify what errors you want to handle, e.g. `except OSError`. This has apparently slipped through review in #46.

mantatail.py
ThePhilgrim: I now removed "( ... )" from the quit reasons.

I am, however, unsure how I'd go about just getting the "title" of a Python error to add that as a quit reason.

mantatail.py
ThePhilgrim: When I added Optional, I thought that would be the case when User didn't give a reason. Then I did it differently. Removed now.

mantatail.py
ThePhilgrim: I'm not always sure what type of error I might expect. Is this `except` protecting against OSError? Is there a good way to manually raise these so I can see what error can happen here?

ThePhilgrim:
@Akuli When making tests, is there any way for me to asserting that a user doesn't receive anything?

I've tried something like `assert not receive_line(user_alice)`, but it still just times out.
How can achieve something that passes if nothing is received?

mantatail.py
Akuli: Sockets always raise `OSError`. I can explain in more detail why that is if you're interested, but what matters for you is that `OSError` is the only error they can raise.

```
>>> s = socket.socket()
>>> s.close()
>>> s.connect(('localhost', 1234))
OSError: [Errno 9] Bad file descriptor
```

But how about other socket errors you have seen, like `ConnectionRefusedError`?

```
>>> socket.socket().connect(('localhost', 1234))
ConnectionRefusedError: [Errno 111] Connection refused
```

These are `OSError`s too. If you catch `OSError`, you will catch these too.

```
>>> try:
...     socket.socket().connect(('localhost', 1234))
... except OSError as e:
...     print("Caught OSError: " + str(e))
... 
Caught OSError: [Errno 111] Connection refused
```

mantatail.py
Akuli: Here are a couple solutions. I'm not sure which of them you call the "title", but pick whichever is what you meant.

```
>>> import socket
>>> try:
...     socket.socket().connect(('localhost', 1234))
... except OSError as e:
...     print("str(e) -->", str(e))
...     print("e.strerror -->", e.strerror)
...     print("e.errno -->", e.errno)
... 
str(e) --> [Errno 111] Connection refused
e.strerror --> Connection refused
e.errno --> 111
```

Akuli:
>How can achieve something that passes if nothing is received?

There is a function that does this. It's called `select.select()`, and dabeaz uses it later in the video I linked a while ago. **But:**

Sending can happen with a delay. So if you use `select.select()` to check if something was sent, your tests can pass even though the server sends something, if there just happens to be a small delay because of how the server's internal threads work.

I can think of a couple solutions that account for the possible delay:
- Sleep a little bit, and then check if the server sent something with `select.select`.
- Use a combination of `receive_line` and `pytest.assert_raises`, so that the test will **pass** if `receive_line` errors. Consider passing e.g. `timeout=0.1` to your `receive_line()` call.

mantatail.py
```diff
@@ -179,6 +181,7 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
                         commands.handle_pong(state, user, args)
                     else:
                         if command_lower == "quit":
+                            disconnect_reason = "Remote host closed the connection"
```
Akuli, line 184: Shouldn't this be "Client quit"?

mantatail.py
```diff
@@ -257,15 +260,15 @@ def send_queue_thread(self) -> None:
         while True:
             (message, prefix) = self.send_que.get()
 
-            if message is None or prefix is None:
+            if message is None:
+                disconnect_reason = prefix
+                quit_message = f"QUIT :Quit: ({disconnect_reason})"
```
Akuli, line 265: These parentheses are going to show up weirdly in mantaray. It adds its own parentheses too, so this will look like:

```
* Alice quit. (Quit: (Connection reset by peer))
```

where the inner parentheses are added by mantatail, and outer by mantaray.

tests/test_mantatail.py
```diff
@@ -658,6 +658,58 @@ def test_quit_before_registering():
...
+    nc4.sendall(b"USER nc4 0 * :netcat\n")
+    nc4.sendall(b"JOIN #foo\n")
+
+    while receive_line(nc4) != b":mantatail 366 nc4 #foo :End of /NAMES list.\r\n":
+        pass
+
+    time.sleep(0.1)
+
+    nc.sendall(b"QUIT\n")
+    assert receive_line(nc4) == b":nc!nc@127.0.0.1 QUIT :Quit: (Client quit.)\r\n"
```
Akuli, line 701: Shouldn't this also check what the quitter himself (`nc`) receives?

mantatail.py
ThePhilgrim: Yeah, you're right!

tests/test_mantatail.py
ThePhilgrim: It does on line 709 when `nc4` quits.

```
nc4.sendall(b"QUIT Many words but no colon\n")
    assert receive_line(nc4) == b":nc4!nc4@127.0.0.1 QUIT :Quit: Many\r\n"
```

I felt it was enough to check what the quitter receives once.

ThePhilgrim:
Lol, I have to go out now so I'll debug this later. But all tests pass on my computer. If I change the test locally to fit the output on github actions, it fails locally

Skärmavbild 2022-01-28 kl  14 12 55

On github actions, the netcat user receives "Remote host closed the connection". When I do ctrl+f on my code, there are no matches on "Remote host closed the connection" in any file.

I'm sure it's simple to debug, but I need to do it later when I have more time.

ThePhilgrim:
Also, when reading the code now. All the quits that happen in `recv_loop()` that jumps to finally, why don't I just call `handle_quit()` from them so that the function gets to handle all quits.

Akuli:
The tests are failing only on Windows, and passing on macos and linux. This is because `strerror` returns a platform-specific error message. If you do something like this, the tests will pass:

```
if sys.platform == "win32":
    assert receive_line(nc) == b":nc!nc@... QUIT :Quit: Remote host closed connection\r\n"
else:
    assert receive_line(nc) == b":nc!nc@... QUIT :Quit: blah blah blah\r\n"
```

But on Windows, it's actually even worse, because the error message depends on what language packs you have installed. For example, I can't get my Windows VM to output `strerror` in english at all. So this would be even better:

```
if sys.platform == "win32":
    # error message depends on windows language
    assert receive_line(nc).startswith(b":nc!nc@... QUIT :Quit: ")
else:
    assert receive_line(nc) == b":nc!nc@... QUIT :Quit: blah blah blah\r\n"
```

Akuli:
> why don't I just call handle_quit() from them so that the function gets to handle all quits.

Like other `handle_` functions, `handle_quit()` is meant to handle commands sent by the user. If the user is getting disconnecting for some other reason than a QUIT message, it conceptually doesn't make sense to call a `handle_` function. But if you think it is a good idea because it's more practical, or it's easier, or it solves a problem etc, just do it :)

ThePhilgrim:
Pushed the merge to get more output in the tests (github actions now runs pytest in verbose mode)

ThePhilgrim:
Ah, in fact, now I see that the output in github actions was not "Remote host closed the connection".

```
            -  b':nc!nc@127.0.0.1 QUIT :Quit: Connection reset by peer\r\n',
E         +  b':nc!nc@127.0.0.1 QUIT :Quit: An existing connection was forcibly closed by t'
E         +  b'he remote host\r\n',
```

The -vv PR was a great addition.

ThePhilgrim:
@Akuli Hopefully, this is ready for a final review.

Akuli (approved):

PR #105, opened 2022-01-27
Away status

ThePhilgrim:
Away functionality based on Libera.

### DIFF
--- a/mantatail.py
+++ b/mantatail.py
@@ -214,6 +214,7 @@ class UserConnection:
 
     Usually the nick is used when referring to the user.
 
+
     Send Queue:
         A send queue and a separate thread are used for sending messages to the client.
         This helps with error handling, and even if someone has a slow internet connection,
@@ -232,7 +233,7 @@ class UserConnection:
         self.nick = "*"
         self.user_message: Optional[List[str]] = None  # Ex. AliceUsr 0 * Alice
         self.user_name: Optional[str] = None  # Ex. AliceUsr
-        self.away: Tuple[bool, str] = (False, "")  # str = away message
+        self.away: Optional[str] = None  # None = user not away, str = user away
         self.send_que: queue.Queue[Tuple[str, str] | Tuple[None, None]] = queue.Queue()
         self.que_thread = threading.Thread(target=self.send_queue_thread)
         self.que_thread.start()
--- a/mantatail.py
+++ b/mantatail.py
@@ -231,6 +231,7 @@ class UserConnection:
         self.nick = "*"
         self.user_message: Optional[List[str]] = None  # Ex. AliceUsr 0 * Alice
         self.user_name: Optional[str] = None  # Ex. AliceUsr
+        self.away: Tuple[bool, str] = (False, "")  # str = away message
         self.send_que: queue.Queue[Tuple[str, str] | Tuple[None, None]] = queue.Queue()
         self.que_thread = threading.Thread(target=self.send_queue_thread)
         self.que_thread.start()
--- a/commands.py
+++ b/commands.py
@@ -188,6 +188,27 @@ def handle_nick(state: mantatail.ServerState, user: mantatail.UserConnection, ar
             state.connected_users[user.nick.lower()] = user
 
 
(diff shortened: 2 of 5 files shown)


mantatail.py
```diff
@@ -232,6 +232,7 @@ def __init__(self, state: ServerState, host: str, socket: socket.socket):
         self.nick = "*"
         self.user_message: Optional[List[str]] = None  # Ex. AliceUsr 0 * Alice
         self.user_name: Optional[str] = None  # Ex. AliceUsr
+        self.away: Tuple[bool, str] = (False, "")  # str = away message
```
Akuli, line 235: How it currently works:
- `(False, "blah")` = impossible
- `(False, "")` = not away
- `(True, "blah")` = away
- `(True, "")` = away, with empty message

I don't like "impossible" cases like this, because what usually happens is that some of the code assumes it doesn't happen, and years later, someone adds code that makes it happen anyway. How about simply `Optional[str]`?

- `None` = not away
- `"blah"` = away
- `""` = away, with empty message

Akuli (commented):
In ircdocs.horse, they describe a way for IRC clients to say "I want to be notified when someone is away/back". This would be a useful feature, and it would let me display away-ness in mantaray's user list.

mantatail.py
ThePhilgrim: On Libera, you cannot set Away status with empty message
`AWAY :Lol` & `AWAY Lol` will both set a user to away.
`AWAY` & `AWAY :` will both set a user to not away.

However, I can absolutely change the Tuple to `Optional[str]` if you think that's better.

I will look into the functionality about showing other users' away messages tomorrow!

Akuli: Ah, so there are two "impossible" cases :D Indeed, `Optional[str]` makes more sense IMO.

ThePhilgrim:
> In ircdocs.horse, they describe a way for IRC clients to say "I want to be notified when someone is away/back". This would be a useful feature, and it would let me display away-ness in mantaray's user list.

@Akuli The only info I seem to find about this is [here](https://ircv3.net/specs/extensions/away-notify.html) (linked from Horse docs). However, this doesn't tell me (and neither does Horse Docs) what command is used for doing this.

The alternative would be to just send the AWAY message to everyone, and then the client handles if these are received or not (depending on the user settings). This would, however, deviate from how Libera is working. On Libera, the away message is not sent to other users.

I do like the client feature of showing who is away (making their nicks a different shade), so if we can find a good way to implement this on the server, I'm all for it!

Akuli:
ircdocs.horse says:

> If the IRCv3 away-notify capability has been requested by a client, the server MAY also send that client AWAY messages to tell them how the away status of other users has changed.

If you Ctrl+F capability, you will find a section named "Capability Negotiation" that explains the concept, and links to https://ircv3.net/specs/extensions/capability-negotiation.html where it is specified in detail.

Mantaray already has some example code where it deals with server capabilities. If you specify a password, mantaray will ask the server whether it supports a capability named `sasl` (`CAP REQ sasl`), and if the server says yes (`CAP * ACK :sasl`), mantaray begins authenticating the user (`AUTHENTICATE PLAIN`). Mantaray is also somewhat prepared for the server to say no (`CAP * NAK :sasl`). Once the login is done, mantaray sends `CAP END`. To see this in action, you can run mantaray with `--verbose` and connect to libera with a password. You can find the relevant code with e.g. `git grep CAP`.

ThePhilgrim:
@Akuli I've now read a lot about capability negotiation and how it works. However, it's a large and rather complicated feature to implement into mantatail. Therefore, it cannot be done in this PR. 

It is a major feature though, and definitely something that I should focus on very soon.

For now, I believe it's best to keep the default functionality to the standard IRC protocol where away messages are _not_ sent to other users.

When `CAP` has been implemented, I can also implement `away-notify`.

What do you think about this?

Akuli (approved):
This looks reasonable. We can now add `/away` and `/back` to mantaray (and it should be much easier than it was a few weeks ago), although we can't make users in the userlist gray yet.

PR #109, opened 2022-01-30
Ban user

ThePhilgrim:
Still missing one test (sending the same flag several times).

I currently see one vulnerability in this PR.

The ban list is based on `UserConnection`.
This means that if the user disconnects, and then connects again. They are no longer banned. This kind of defeats the whole purpose of banning. 

The test currently marked with xfail (that actually passes) is not supposed to make it into main. But it covers a question I have. When a banned user disconnects, and then a user on the channel asks for the ban list `MODE #chan +b`, mantatail fetches the ban list (which is based on `UserConnection`). The ban list message sends `user.nick`, but that UserConnection doesn't exist anymore. Shouldn't this error? Or does it NOT error because the UserConnection is never removed from the ban list?

 It seems that usually, ban lists are based on user masks, but with wildcards, ex `Usernick!*@*`.
That would require me to store it as a string. These types of issues are going to remain until the server supports proper registration.
One solution would be to remove `UserConnection` from the ban list when they disconnect. But we would have to make sure this is changed in the future so that things can be stored across reconnects.

Would love to hear your thoughts on this, and the PR in general to see what I can improve.


Akuli:
> It seems that usually, ban lists are based on user masks, but with wildcards, ex `Usernick!*@*`. That would require me to store it as a string. These types of issues are going to remain until the server supports proper registration.

I don't see why proper registration is needed for this? An operator should be able to ban anyone, registered or not. Especially the part after `@` is useful, because that is the IP or hostname that the user is connecting from (such as `127.0.0.1`).

Speaking of wildcards, the `fnmatch` module might work for checking if a user's mask matches a string containing wildcards.

> things can be stored across reconnects

It would be pretty bad to get trolls rejoining your channel just because someone had to restart the server. How about just saving the ban lists to a JSON file for now?

ThePhilgrim:
I'm talking about registering because I get the feeling that will tie a user more to their user mask. They are identified with their user account.

If there is no registering -> person A connects with `nick Foo` -> `user Foo 0 * BarBaz`, then that person quits.. and then another person B on the other side of the world connects with `nick Foo` -> `user Foo 0 * BarBaz`, then that person will be banned from the channels person A got banned from.

If person A and person B are registered, this wouldn't be a problem?

Akuli:
> person will be banned from the channels person A got banned from

When banning someone, you decide whether you want to based on nick, user or IP. Each kind of ban is different, and has its own use:
- Nick ban: Good if someone's nickname is offensive. Anyone else with the same nick will be banned too. You can work around this by changing to a less offensive nick.
- User ban: Mostly useless unless the user is registered. But even then, easy to work around by not using your registered account.
- IP ban: Use this if you just want to get rid of a person, or a botnet of computers that are connecting through the same internet connection. Bans everyone who use an internet connection, at least until their IP changes for whatever reason.

This is just how IRC works. When you ban someone, it's your job to figure out how to identify the user to ban. The IRC server can't help much with it anyway.

Also, I have never needed to actually ban someone, so I might have gotten some detail wrong :)

ThePhilgrim:
This is my current pseudo code for the ban function:

```
def generate_ban_mask(ban_target: str) -> str:
    # 
    # if "!" in ban_target and "@" in ban_target:
    #   regex for *!*@*
    #   regex_groups_list = blah.groups()
    # elif "!" in ban_target:
    #   regex for *!*
    #   regex_groups_list = blah.groups()
    # elif "@" in ban_target:
    #   regex for *@*
    #   regex_groups_list = blah.groups()
    # else:
    #   regex for *
    #   regex_groups_list = blah.groups()
    #
    # final_mask = []
    #
    # for x in regex_groups_list:
    #   if not x:
    #     final_mask.append correct thingy (! or @)
    #   else:
    #     final_mask.append x
```

Anything I need to consider? That I missed etc?

ThePhilgrim:
@Akuli 
With the new logic (ban masks), the tests are currently failing.
The reason for this is that I need to find a way to match the `MODE` command parameters with what is already in the `ban_list`.

I've looked at `fnmatch`, but it doesn't seem like it will be the solution. The problem with this is that you can only use wildcards (`*`& `?`) on one side. See screenshot:

Skärmavbild 2022-02-02 kl  15 53 48

This does of course become very problematic since the `*` in the ban list will not be considered a wildcard, but a literal asterisk character.

The only solution I can think of is to loop through the ban list, and if the mask in the list has a `*`, replace whatever is on that position in the `+b` parameter.

Ex. 
Command: `MODE #chan +b foo!bar@baz`
Ban List: `["foo!*@*]`

Split them both:
Command parameter: `["foo", "!", "bar", "@", "baz"]`
Ban List: `["foo", "!", "*", "@", "*"]`

Compare, replace `bar` with `*` & `baz` with `*`

That would then match.

But it's a horribly annoying and inelegant solution. Do you have any better ideas?

Akuli:
Something like this?

```
>>> ban_list = ["foo", "b*r", "ba*"]
>>> any(fnmatch.fnmatch("bar", ban_wildcard) for ban_wildcard in ban_list)
True
>>> any(fnmatch.fnmatch("Akuli", ban_wildcard) for ban_wildcard in ban_list)
False
```

ThePhilgrim:
Hmm, it still doesn't seem to work out so well.
Skärmavbild 2022-02-02 kl  18 56 18

Akuli:
When you write `fnmatch.fnmatch("*!bar@baz", ban_wildcard)`, you're doing it wrong. The first argument to `fnmatch.fnmatch` should be an actual user mask, not a wildcard. The second argument is what contains the wildcards.

ThePhilgrim:
@Akuli I thought implementing the ban feature would be super simple. Instead, it proved to be quite complicated.

I believe that it is now ready for a final review.
I decided to follow Libera's functionality where `foo@bar!baz` automatically gets changed to `foo!bar@baz`.

commands.py
```diff
@@ -386,6 +396,9 @@ def handle_privmsg(state: mantatail.ServerState, user: mantatail.UserConnection,
 
     Depending on the command, sends a message to all users on a channel or a private message to a user.
     """
+
+    # TODO: Check if user is in channel ban list
```
Akuli, line 400: This seems to be done?

commands.py
```diff
@@ -110,6 +112,14 @@ def handle_join(state: mantatail.ServerState, user: mantatail.UserConnection, ar
 
         assert channel
 
+        match_with_banned_user = any(
+            fnmatch.fnmatch(user.get_user_mask(), ban_mask) for ban_mask in channel.ban_list.keys()
+        )
```
Akuli, line 117: This code is repeated in a couple places. How about making a method for it? E.g. `user.is_banned_from(channel)` or `channel.has_ban_for(user)`

commands.py
```diff
@@ -491,45 +513,110 @@ def process_channel_modes(state: mantatail.ServerState, user: mantatail.UserConn
 
     if len(args) == 1:
         if channel.modes:
-            message = f'{irc_responses.RPL_CHANNELMODEIS} {user.nick} {channel.name} {" ".join(channel.modes)}'
+            message = f'{irc_responses.RPL_CHANNELMODEIS} {user.nick} {channel.name} +{" ".join(channel.modes)}'
```
Akuli, line 516: This doesn't seem correct to me. If modes `a` and `b` are sent, this sends `+a b`, with `+` only in front of `a`.

commands.py
```diff
@@ -491,45 +513,110 @@ def process_channel_modes(state: mantatail.ServerState, user: mantatail.UserConn
...
         user.send_que.put((message, "mantatail"))
-    elif len(args) == 2:
-        error_not_enough_params(user, "MODE")
     else:
         if args[1][0] not in ["+", "-"]:
             error_unknown_mode(user, args[1][0])
             return
-        supported_modes = ["o"]
+
+        valid_chanmodes = r"[a-zA-Z]"
```
Akuli, line 525: I don't see a reason to have this in a variable. I think `r"[a-zA-Z]"` by itself would be more readable than `valid_chanmodes`.

commands.py
```diff
@@ -491,45 +513,110 @@ def process_channel_modes(state: mantatail.ServerState, user: mantatail.UserConn
...
+    if user not in channel.operators:
+        error_no_operator_privileges(user, channel)
+        return
 
-                if mode_command == "+":
-                    channel.operators.add(target_usr)
-                elif mode_command[0] == "-":
-                    channel.operators.discard(target_usr)
+    target_ban_mask = generate_ban_mask(ban_target)
+    match_with_banned_user = any(fnmatch.fnmatch(target_ban_mask, ban_mask) for ban_mask in channel.ban_list.keys())
```
Akuli, line 574: This doesn't seem right to me. You are passing two wildcarded strings to `fnmatch.fnmatch`. One of them should be wildcarded and the other shouldn't.

commands.py
```diff
@@ -562,6 +649,62 @@ def parse_received_args(msg: str) -> Tuple[str, List[str]]:
...
+    """
+    Generates a user mask based on the parameters given in a MODE +b command.
+    Any part of the user mask not provided by the user is added as a wildcard ("*").
+
+    Ex:
+        - MODE +b Foo -> Foo!*@*
+        - MODE +b Foo!Bar -> Foo!Bar@*
+        - MODE +b Foo!Bar@Baz -> Foo!Bar@Baz
+        - MODE +b Bar@Baz -> *!Bar@Baz
+        - MODE +b @Baz -> *!*@Baz
```
Akuli, line 662: This would be a good use case for doctests. Google something like `python doctest` if you aren't familiar with the concept, and `pytest doctest` for how to use doctests in projects that use pytest.

commands.py
```diff
@@ -562,6 +649,62 @@ def parse_received_args(msg: str) -> Tuple[str, List[str]]:
...
+
+        assert ban_match is not None  # Keeps mypy silent
+        nick, user, host = ban_match.groups()
+
+        if not nick:
+            nick = "*"
+        if not user:
+            user = "*"
+        if not host:
+            host = "*"
```
Akuli, line 680: These are repeated a lot. I would probably put them after the if statements, so that they are applied in all cases.

tests/fuzzer/fuzzer.py
```diff
@@ -29,9 +29,11 @@
     "#foo",
     "+g",
     "+o",
+    "+b",
     "+xyz",
     "-g",
     "-o",
+    "-b",
```
Akuli, line 36: We could probably add some user masks here too.

tests/test_mantatail.py
```diff
@@ -426,7 +426,88 @@ def test_channel_mode_is(user_alice):
         pass
 
     user_alice.sendall(b"MODE #foo\r\n")
-    assert receive_line(user_alice) == b":mantatail 324 Alice #foo\r\n"
+    assert receive_line(user_alice) == b":mantatail 324 Alice #foo +t\r\n"
```
Akuli, line 429: What mode is `+t`?

commands.py
Akuli:
```suggestion
            message = f'{irc_responses.RPL_CHANNELMODEIS} {user.nick} {channel.name} +{"".join(channel.modes)}'
```

mantatail.py
```diff
@@ -375,9 +383,10 @@ class Channel:
     def __init__(self, channel_name: str, user: UserConnection) -> None:
         self.name = channel_name
         self.topic: Optional[Tuple[str, str]] = None  # (Topic, Topic author)
-        self.modes: List[str] = []
+        self.modes: Set[str] = {"t"}
```
Akuli, line 386:
```suggestion
        self.modes: Set[str] = {"t"}  # t = only operator can set topic
```

Akuli (approved):

PR #110, opened 2022-01-30
Capability Negotiation

ThePhilgrim:
This is a start on implementing capability negotiation.

Still need to:
- [x] Implement various tests
- [x] Add commands to fuzzer
- [x] Make Mantatail wait for `CAP END`

@Akuli 
I would need your help to understand the best way to make the server wait for `CAP END`.

If the user sends:
```
CAP LS 302
NICK foo
USER foo 0 * bar
```

The server does not, at this point, send MOTD and other messages. The server silently waits until it receives `CAP END`. Then it sends appropriate welcome messages.

Note, however, that I cannot just set a bool like `capneg_in_progress = True` and add a simple if statement in `recv_loop()`. Because then the MOTD would not get sent when `CAP END` is finally sent.

I guess I can do something more complicated with bools, like `capneg_initiated = True`, `capend_sent = False`, then set the latter to `True`. But that becomes really complicated, because then I need to do that check -> accept other commands, but for MOTD I also need to set a bool `motd_sent = True` to not send it over and over.. 

It becomes a whole lot of bool checks.
Is there a more elegant way to solve this?


mantatail.py
```diff
@@ -24,13 +24,18 @@ class ServerState:
...
+            to iterables during iteration.
+
+            - cap_ls: Server Capabilities. List of features beyond the standard
+            IRC protocol that Mantatail supports.
         """
 
         self.lock = threading.Lock()
         self.channels: Dict[str, Channel] = {}
         self.connected_users: Dict[str, UserConnection] = {}
+        self.cap_ls: List[str] = ["cap-notify", "foo", "bar", "baz"]
```
Akuli, line 38: Why is this in the state? Is it meant to be changed while the server runs? Why not a global variable, similar to `irc_responses.py`?

Also, why are `"foo"`, `"bar"` and `"baz"` here?

mantatail.py
```diff
@@ -234,12 +241,13 @@ def __init__(self, state: ServerState, host: str, socket: socket.socket):
...
         self.nick = "*"
-        self.user_message: Optional[List[str]] = None  # Ex. AliceUsr 0 * Alice
-        self.user_name: Optional[str] = None  # Ex. AliceUsr
+        self.user_message: Optional[List[str]] = None
+        self.user_name: Optional[str] = None
         self.away: Optional[str] = None  # None = user not away, str = user away
         self.send_que: queue.Queue[Tuple[str, str] | Tuple[None, str]] = queue.Queue()
         self.que_thread = threading.Thread(target=self.send_queue_thread)
         self.que_thread.start()
+        self.cap_list: Set[str] = set()
```
Akuli, line 250: I don't like the name. A person new to mantatail would look at `cap_list` and think it's a list (wrong), wondering what a cap is (bad). How about `capabilities`? It shouldn't be `capabilities_set` though, because it isn't hard to tell what type it is (hover or ctrl+click in most editors).

commands.py
```diff
@@ -27,6 +27,59 @@
...
+
+    if low_cap_command == "req":
+        # CAP REQ without specified capability
+        if len(args) == 1:
+            return
+
+        capabilities = args[1].split(" ")
+        unsupported_caps = [cap for cap in capabilities if cap not in state.cap_ls]
+
+        if unsupported_caps:
```
Akuli, line 68: What should happen if some capabilities are supported and others are not? Should it send both ACK and NAK, or only NAK? Should the unsupported capabilities be enabled?

mantatail.py
ThePhilgrim: `cap_ls` will never be changed while the server is running. However, I would say supported modes are still a state of the server. In the future, it will be updated with more features, which then would be a different state of the server. I think that makes sense. Don't you?

foo bar baz was used while testing, and should not have been included in the commit. Will remove it in the next push.

mantatail.py
ThePhilgrim: "list" here has nothing to do with the python data structure. It is named, just like `cap_ls` after the specific `CAP` command.

`CAP LIST` shows what features have been negotiated (enabled) by the client. 
I could change the name to `enabled_capabilities`, but then `cap_ls` has to be renamed as well.

cap_list (or whatever it will be named) will be described in the docstring anyways, so a developer can find any necessary information there.

commands.py
ThePhilgrim: This is designed in accordance with Libera.
If 3 capabilities are `REQ`'ed in the same command, and 1 is unsupported, it gives `NAK` for the whole command, and nothing is added to `CAP LIST`.

Akuli:
> Note, however, that I cannot just set a bool like `capneg_in_progress = True` and add a simple if statement in recv_loop(). Because then the MOTD would not get sent when CAP END is finally sent.

I don't understand why this is a problem. Currently it looks like this (in pseudo code):

```
if command is nick:
    nick_received = True
if command is user:
    user_received = True

if nick_received and user_received:
    send_motd()
```

Why couldn't it be something like this:

```
if command is nick:
    nick_received = True
if command is user:
    user_received = True
if command is CAP END:
    capneg_in_progress = False

if nick_received and user_received and not capneg_in_progress:
    send_motd()
```

Because sending `CAP END` can trigger motd, we can put it in the same place where we handle NICK and USER. Shouldn't this work?

mantatail.py
Akuli: > supported modes are still a state of the server

What exactly is "the server" here? A process running on someone's computer (`State` makes sense to me too), or the project as a whole (a global variable would be better IMO)?

mantatail.py
```diff
@@ -169,26 +178,31 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
...
                     if command_lower == "user":
                         if args:
                             user.user_message = args
                             user.user_name = args[0]
+                            user_received = True
                         else:
                             commands.error_not_enough_params(user, command)
                     elif command_lower == "nick":
                         commands.handle_nick(state, user, args)
+                        nick_received = True
```
Akuli, line 191: Try sending `NICK` with no arguments followed by `USER foo`. You will notice a problem :)

Basically, the same data is in two places: `user.nick` can be `*` or something else, and `nick_received` can be `False` or `True`.

ThePhilgrim: Ah, well spotted. So instead of `nick_received`, I'll use `if user.nick != "*"`. Right?

Akuli: That would make sense to me, and it works that way on main.

ThePhilgrim: Perfect!

ThePhilgrim:
@Akuli I think this PR is ready for a final review.

When this is merged, `away-notify` should be like a 1-5 line fix!

mantatail.py
```diff
@@ -15,22 +15,27 @@
...
     def __init__(self, motd_content: Optional[Dict[str, List[str]]]) -> None:
         """
-        The attribute "self.lock" locks the state of the server to avoid modifications
-        to iterables during iteration.
+        Attributes:
+            - lock: Locks the state of the server to avoid modifications
+            to iterables during iteration.
+
+            - cap_ls: Server Capabilities. List of features beyond the standard
+            IRC protocol that Mantatail supports.
```
Akuli, line 32: Please revert the docstring changes. Nothing in this class needed changing after all.

tests/fuzzer/fuzzer.py
```diff
@@ -33,6 +33,10 @@
     "-g",
     "-o",
     "-xyz",
+    "CAP",
+    "LS",
+    "LIST",
+    "REQ",
```
Akuli, line 39:
```suggestion
    "REQ",
    "END",
```

tests/test_mantatail.py
```diff
@@ -691,14 +691,73 @@ def test_connect_via_netcat(run_server):
...
+
+def test_quit_before_registering(run_server):
     with socket.socket() as nc:
         nc.connect(("localhost", 6667))  # nc localhost 6667
         nc.sendall(b"QUIT\n")
         assert receive_line(nc) == b":QUIT :Quit: Client quit\r\n"
 
 
-def test_quit_reasons():
+def test_quit_reasons(run_server):
```
Akuli, line 760: You can delete `autouse=True` from the `run_server` fixture, if all tests use it explicitly.

Akuli (approved):

PR #114, opened 2022-02-04
Code Split Up

ThePhilgrim:


mantatail.py
```diff
@@ -157,72 +157,119 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
...
 
-    user = UserConnection(state, user_host, user_socket)
-    motd_sent = False
-    disconnect_reason = ""
+    def __init__(self, state: ServerState, user_host: str, user_socket: socket.socket) -> None:
+        self.state = state
+        self.user_host = user_host
+        self.user_socket = user_socket
+        self.user = UserConnection(state, user_host, user_socket)
+        self.motd_sent: bool = False
```
Akuli, line 165: Does `motd_sent` really need to be saved to `self`? It seems to be used only inside `receive_loop()`.

mantatail.py
```diff
@@ -157,72 +157,119 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
...
+                        else:
+                            with self.state.lock:
+                                call_handler_function(self.state, self.user, args)
+                                if command_lower == "quit":
+                                    return
+
+        finally:
+            self.user.send_que.put((None, self.disconnect_reason))
+
+    def get_message_request(self) -> bytes | None:
```
Akuli, line 214: If I look at the function name, I'm screwed, because I have no idea what a message request is. (IRC doesn't even have a thing called a "request"; that's more of a HTTP thing.) But if I look at how the method is implemented, I see `while not something.endswith(b"\n")` and I can easily guess what it does.

I think the method needs to be renamed. How about `receive_lines()`?

mantatail.py
```diff
@@ -157,72 +157,119 @@ def recv_loop(state: ServerState, user_host: str, user_socket: socket.socket) ->
...
-                if request_chunk:
-                    request += request_chunk
-                else:
-                    disconnect_reason = "Remote host closed the connection"
+    def recv_loop(self) -> None:
+        try:
+            while True:
+                request = self.get_message_request()
+
+                if not request:
```
Akuli, line 175: For future reference: `if not bytes_or_none:` is potentially confusing, because it treats `None` and empty bytes object in the same way. An explicit `if thingy is not None` is better, if that's what you want to do. But in this case, it doesn't matter, because `get_message_request()` can't return empty bytes.

ThePhilgrim:
Currently, the tests pass locally, but mypy is angry about the imports.

If I just do `import errors` etc, mypy is happy, but tests error.

I've tried:
`from mantatail import foo, bar`

ThePhilgrim:
Regarding the helper functions, I decided to opt for the solution with `class Helpers`. 
Using the solution with nested functions didn't work for me. I could do `assert receive_line...` but not `while receive_line...` as fixtures "are not supposed to be called directly."

server.py
```diff
@@ -13,14 +13,13 @@
 import json
```
Akuli, line 13: Because this file is renamed, you now have to run `python3 server.py` if you want to start mantatail. Please update the README accordingly.

server.py
```diff
@@ -113,116 +112,164 @@ def run_server_forever(self) -> None:
...
+                                call_handler_function(self.state, self.user, args)
+                                if command_lower == "quit":
+                                    return
+
+        finally:
+            self.user.send_que.put((None, self.disconnect_reason))
+
+    def receive_messages(self) -> bytes | None:
+        """
+        Receives bytes from the client and returns them to recv_loop().
```
Akuli, line 201: I think you should mention that it receives entire lines, possibly multiple lines at once.

server.py
```diff
@@ -320,6 +367,7 @@ def queue_quit_message_for_other_users(self, quit_message: str) -> None:
             receiver.send_que.put((quit_message, self.get_user_mask()))
 
     def get_users_sharing_channel(self) -> Set[UserConnection]:
+        """Returns a set of all users sharing a channel with the client."""
```
Akuli, line 370: You are repeating what the method name already says. As it is currently, the docstring doesn't provide any useful information to the reader. You can delete the docstring, or reword it:

```suggestion
        """Returns all users of all channels that this user has joined."""
```

tests/conftest.py
```diff
@@ -0,0 +1,159 @@
...
+        "- Baz",
+        "-",
+        "- End test MOTD"
+        ]
+}
+
+# fmt: on
+
+# Based on: https://gist.github.com/sbrugman/59b3535ebcd5aa0e2598293cfa58b6ab#gistcomment-3795790
+@pytest.fixture(scope="function")
```
Akuli, line 24: `scope="function"` is the default, and doesn't need to be specified:

```suggestion
@pytest.fixture
```

tests/conftest.py
```diff
@@ -0,0 +1,159 @@
...
+                nonlocal last_exception
+                last_exception = e
+
+    monkeypatch.setattr(threading, "Thread", ThreadWrapper)
+    yield
+    if last_exception:
+        raise last_exception
+
+
+@pytest.fixture()
```
Akuli, line 43: It's confusing that this one has `()` at the end while other fixtures don't. It works the same either way, but consistency would be good IMO.

```suggestion
@pytest.fixture
```

tests/conftest.py
```diff
@@ -0,0 +1,159 @@
...
+    yield charlie_socket
+    charlie_socket.sendall(b"QUIT\r\n")
+    while b"QUIT" not in helpers.receive_line(charlie_socket):
+        pass
+    charlie_socket.close()
+
+
+# Based on https://stackoverflow.com/a/42156088/15382873
+class Helpers:
+    def receive_line(sock, timeout=1):
```
Akuli, line 124: Because these methods don't take `self`, it would be good style to add `@staticmethod` to them. But it doesn't really matter because you never instantiate the class.

tests/conftest.py
```diff
@@ -0,0 +1,159 @@
...
+# @pytest.fixture
+# def receive_line():
+#     def execute_receive_line(sock, timeout=1):
+#         sock.settimeout(timeout)
+#         received = b""
+#         while not received.endswith(b"\r\n"):
+#             received += sock.recv(1)
+#         return received
+
+#     return execute_receive_line
```
Akuli: If you want, we can make these fixtures work and delete `class Helpers`. If you don't, delete these comments

ThePhilgrim:
@Akuli I think I've split the code as much as makes sense or is possible. What do you think?

README.md
```diff
@@ -35,6 +45,16 @@ To start, please follow these steps:
 
 Mantatail relies on Pytest for testing. In order to run the test suite, use `python3 -m pytest`
 
+### Fuzzing
+
+The fuzzer spams the server with random commands to reveal eventual errors.
+To run the fuzzer, make sure `server.py` is not running, cd to the mantatail folder and run `python3 fuzzer.py`
```
Akuli, line 51:
```suggestion
To run the fuzzer, make sure `server.py` is not running, cd to `tests/fuzzer` and run `python3 fuzzer.py`.
```

README.md
```diff
@@ -35,6 +45,16 @@ To start, please follow these steps:
...
+
+If you are developing a new feature, you can substitute the list `words` with a list containing limited amount of relevant data.
+
+For example, if you are working with the feature of banning a user, a reasonable data set could be:
+`words = ["MODE", "mode", "Mode", "#chan", "#%invalidchan "+", "-", "b", "B", "^", ":", "Param1", ":Param2", "Param with spaces"]`
```
Akuli, line 56: Spaces aren't needed, because the fuzzer joins the words with spaces.
```suggestion
`words = ["MODE", "mode", "Mode", "#chan", "#%invalidchan "+", "-", "b", "B", "^", ":", "Param1", ":Param2"]`
```

tests/conftest.py
```diff
@@ -0,0 +1,139 @@
...
+    @staticmethod
+    def receive_line(sock, timeout=1):
+        sock.settimeout(timeout)
+        received = b""
+        while not received.endswith(b"\r\n"):
+            received += sock.recv(1)
+        return received
+
+    # Makes it easier to assert bytes received from Sets
+    def compare_if_word_match_in_any_order(received_bytes, compare_with):
```
Akuli, line 133:
```suggestion
    @staticmethod
    def compare_if_word_match_in_any_order(received_bytes, compare_with):
```

Akuli:
I like the current amount of splitting :)

README.md
ThePhilgrim: @Akuli You made the fuzzer cd to `tests/fuzzer` by itself, so no need manually `cd`ing there.

Akuli: In that case you need to run `python3 tests/fuzzer/fuzzer.py`.

README.md
ThePhilgrim: Hmm, this might be problematic though. I guess that the fuzzer needs to support something like `PRIVMSG #chan :Param with spaces`, don't you think?

README.md
ThePhilgrim: Ah, okay. Misunderstanding by me then

README.md
Akuli: With `"Param with spaces"` removed, it will send things like `:Param2 Param1`, which is essentially a param with spaces.

ThePhilgrim: Perfect

ThePhilgrim:
Everything resolved. Let's have a look at the tests.

PR #115, opened 2022-02-05
Listen to connections coming from other computers within the same network

Akuli:
Tested manually with two computers.

### DIFF
--- a/server.py
+++ b/server.py
@@ -94,7 +94,7 @@ class ConnectionListener:
     """Starts the server and listens for incoming connections from clients."""
 
     def __init__(self, port: int, motd_content: Optional[Dict[str, List[str]]]) -> None:
-        self.host = "127.0.0.1"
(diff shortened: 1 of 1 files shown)


server.py
```diff
@@ -94,7 +94,7 @@ class ConnectionListener:
     """Starts the server and listens for incoming connections from clients."""
 
     def __init__(self, port: int, motd_content: Optional[Dict[str, List[str]]]) -> None:
-        self.host = "127.0.0.1"
+        self.host = ""
```
ThePhilgrim, line 97: @Akuli, Thank you for this PR! Would you please explain how simply making `self.host` into an empty string allows connections from other computers?

Akuli: That's how `.bind()` works. See the example in docs: https://docs.python.org/3/library/socket.html#example

PR #116, opened 2022-02-05
Registration messages

### LINKED ISSUE #111: Sending supported prefixes with RPL_ISUPPORT
Under "Channel Membership Prefixes", modern.ircdocs.horse says that the prefixes:

> MUST be advertised by the server in the PREFIX RPL_ISUPPORT parameter on connection.

Currently mantatail uses the @ prefix to denote operators, but doesn't send RPL_ISUPPORT.

ThePhilgrim:
~~Still have to add tests~~ Done

Fixes #111


server.py
```diff
@@ -328,6 +334,28 @@ def get_nick_with_prefix(self, channel: Channel) -> str:
...
+
+        The messages sent:
+            - 001 Welcome message (RPL_WELCOME)
+            - 002 Name and version of the server (RPL_YOURHOST)
+            - 003 When the server was created (RPL_CREATED)
+            - 004 All supported user/channel modes (RPL_MYINFO)
+            - 005 Features and limitations of the server (RPL_ISUPPORT)
+
+        This function calls all the appropriate functions to make sure the client receives all appropriate messages.
+        """
```
Akuli, line 350: Try to keep docstrings at a high level (if it mentions "functions", it isn't high level enough), and avoid duplication. IMO the list of replies and  the sentence about calling functions can be deleted. But the first sentence is good :)

server.py
```diff
@@ -11,18 +11,22 @@
...
 from typing import Dict, Optional, List, Set, Tuple
 
 import commands, errors
 
-TIMER_SECONDS = 600
+MANTATAIL_VERSION = "0.0.1"
+SERVER_STARTED = datetime.today().ctime()
+PING_TIMER_SECS = 300
 CAP_LS: List[str] = ["away-notify", "cap-notify"]
+ISUPPORT = {"NICKLEN": "16", "PREFIX": "(o)@", "CHANTYPES": "#", "TARGMAX": "PRIVMSG:1,JOIN:1,PART:1,KICK:1"}
```
Akuli, line 23: What does `(o)` do?

server.py
```diff
@@ -39,6 +43,8 @@ def __init__(self, motd_content: Optional[Dict[str, List[str]]]) -> None:
         self.lock = threading.Lock()
         self.channels: Dict[str, Channel] = {}
         self.connected_users: Dict[str, UserConnection] = {}
+        self.host = host
+        self.port = str(port)
```
Akuli, line 47: Ports are numbers, so storing  them as strings doesn't make sense to me.

```suggestion
        self.port = port
```

commands.py
```diff
@@ -455,6 +455,38 @@ def privmsg_to_user(state: server.State, sender: server.UserConnection, receiver
         sender.send_que.put((away_message, "mantatail"))
 
 
+def rpl_welcome(user: server.UserConnection) -> None:
+    welcome_msg = f"001 {user.nick} :Welcome to Mantatail {user.get_user_mask()}"
+    user.send_que.put((welcome_msg, "mantatail"))
```
Akuli, line 460: These functions are so short that I personally would just write their codes into `on_registration`. But if you like short functions, this is fine :)

commands.py
```diff
@@ -455,6 +455,38 @@ def privmsg_to_user(state: server.State, sender: server.UserConnection, receiver
...
+def rpl_myinfo(user: server.UserConnection, state: server.State) -> None:
+    all_chanmodes_joined = "".join([chanmode for key in state.chanmodes.keys() for chanmode in state.chanmodes[key]])
+    myinfo_msg = f"004 {user.nick} Mantatail {server.MANTATAIL_VERSION} {all_chanmodes_joined}"
+    user.send_que.put((myinfo_msg, "mantatail"))
+
+
+def rpl_isupport(user: server.UserConnection) -> None:
+    isupport = []
+
+    # ISUPPORT = {"NICKLEN": "16", "PREFIX": "(o)@", "CHANTYPES": "#", "TARGMAX": "PRIVMSG:1,JOIN:1,PART:1,KICK:1"}
```
Akuli, line 482:
```suggestion
```

server.py
ThePhilgrim: `(o)` is for operator. Libera, for example, sends `PREFIX=(ov)@+` (operator and voiced)

commands.py
ThePhilgrim: I did this to avoid making `on_registration` do too many things. I consider these messages separate, and therefore, I think they should be in their own functions. 

If there would be a command like `CREATED`, I'd like to be able to call `rpl_created()` instead of having to rewrite the code. I feel this way to work is best for scalability.

tests/test_basic_functionality.py
```diff
@@ -209,6 +209,22 @@ def test_connect_via_netcat(run_server, helpers):
...
+    nc = socket.socket()
+    nc.connect(("localhost", 6667))
+    nc.sendall(b"NICK nc\n")
+    nc.sendall(b"USER nc 0 * :netcat\n")
+
+    assert helpers.receive_line(nc) == b":mantatail 001 nc :Welcome to Mantatail nc!nc@127.0.0.1\r\n"
+    assert b":mantatail 002 nc :Your host is Mantatail[" in helpers.receive_line(nc)
+    assert b":mantatail 003 nc :This server was created" in helpers.receive_line(nc)
+    assert b":mantatail 004 nc Mantatail 0.0.1" in helpers.receive_line(nc)
+    assert b":mantatail 005 nc" and b":are supported by this server" in helpers.receive_line(nc)
```
Akuli, line 222: This doesn't do what you intended. `"foo" and "bar" in baz` means `"foo" and ("bar" in baz)`, so it ends up doing `and` between a string and a boolean.

```
>>> b":mantatail 005 nc" and True
True
>>> b":mantatail 005 nc" and False
False
```

```suggestion
    line = helpers.receive_line(nc)
    assert b":mantatail 005 nc" in line
    assert b":are supported by this server" in line
```

server.py
```diff
@@ -39,6 +43,8 @@ def __init__(self, motd_content: Optional[Dict[str, List[str]]]) -> None:
         self.lock = threading.Lock()
         self.channels: Dict[str, Channel] = {}
         self.connected_users: Dict[str, UserConnection] = {}
+        self.host = host
```
Akuli, line 46: Is this still needed?

PR #117, opened 2022-02-06
Chanmode +/-t

ThePhilgrim:
Allows channel operators to remove (and set) channel mode `t`.

`t` is a channel mode, enabled by default, that allows only channel operators to set or remove channel topics.


commands.py
```diff
@@ -627,11 +626,27 @@ def process_mode_o(
...
-    elif mode_command[0] == "-" and target_usr in channel.operators:
+    elif mode_command == "-" and target_usr in channel.operators:
         channel.queue_message_to_chan_users(mode_message, user)
         channel.operators.discard(target_usr)
 
 
+def process_mode_t(user: server.UserConnection, channel: server.Channel, mode_command: str) -> None:
+    if user not in channel.operators:
+        errors.no_operator_privileges(user, channel)
+        return
```
Akuli, line 637: These lines seem to be untested. There's no tests where a non-operator tries to mess with the `t` flag.

PR #122, opened 2022-02-08
update PING and PONG timers

### LINKED ISSUE #119: PING & PONG timers
Libera seems to have a much longer PING timer than mantatail. This should be set to something like 2-4 minutes.
The PONG timer should be 240 seconds

chandleroking:
resolves #119


ThePhilgrim:
@chandleroking Thank you for your contribution. Much appreciated. If you would like to work on more issues or new features, you are most welcome to do so. I would be happy to indicate some features that would be nice to implement.

PR #130, opened 2022-02-24
WHOIS support

### LINKED ISSUE #128: Implement `RPL_WHOISUSER`
Since https://github.com/Akuli/mantaray/pull/244, Mantaray relies on `311 RPL_WHOISUSER` to save the client's nickmask. Knowing the length of the nickmask helps to calculate the maximum length of messages.

Currently, the Mantaray test for this logic fails when running it on Mantatail. Implement support for `WHOIS` to support these tests.

ThePhilgrim:
This should provide enough support of `WHOIS` for Mantaray tests running in https://github.com/Akuli/mantaray/pull/244.

The functionality is based on Libera, where you can add more than one parameter.

With one parameter, the server checks if the parameter corresponds to an existing user.
With two parameters, it checks if param1 is an existing server. If it's not, the user will get a `402 No such server`.

Fixes #128

### DIFF
--- a/commands.py
+++ b/commands.py
@@ -469,6 +469,24 @@ def handle_who(state: server.State, user: server.UserConnection, args: List[str]
     user.send_que.put((end_of_who_message, "mantatail"))
 
 
+def handle_whois(state: server.State, user: server.UserConnection, args: List[str]) -> None:
+    if not args:
+        errors.not_enough_params(user, "WHOIS")
+        return
+
+    whois_user = None
+
+    if len(args) == 2:
+        whois_user = state.find_user(args[1])
+    else:
+        pass
+        
+
+    if not whois_user:
+        errors.no_such_nick_channel(user, whois_user)
+        ---
+
+
 def handle_pong(state: server.State, user: server.UserConnection, args: List[str]) -> None:
     """
     Handles client's PONG response to a PING message sent from the server.
--- a/commands.py
+++ b/commands.py
@@ -485,7 +485,9 @@ def handle_whois(state: server.State, user: server.UserConnection, args: List[st
     elif len(args) >= 2:
         param_2_user = state.find_user(args[1])
 
-        if not param_1_user and args[0] != "mantatail":
+        if (
+            not param_1_user and args[0] != "mantatail.chat"
+        ):  # TODO: Update "mantatail.chat" when/if server gets an official name.
             errors.no_such_server(user, args[0])
         elif not param_2_user:
             errors.no_such_nick_channel(user, args[1])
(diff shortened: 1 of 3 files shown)


commands.py
```diff
@@ -469,6 +469,31 @@ def handle_who(state: server.State, user: server.UserConnection, args: List[str]
...
+    if len(args) == 1:
+        if not param_1_user:
+            errors.no_such_nick_channel(user, args[0])
+        else:
+            whois_reply(user, param_1_user)
+
+    elif len(args) >= 2:
+        param_2_user = state.find_user(args[1])
+
+        if not param_1_user and args[0] != "mantatail":
```
Akuli, line 488: This might behave weirdly if someone does `/nick mantatail`. Usually server names contain a dot, as in `strontium.libera.chat`. Maybe we should just not implement the two-argument version yet?

errors.py
```diff
@@ -85,6 +85,12 @@ def banned_from_chan(user: server.UserConnection, channel: server.Channel) -> No
     user.send_que.put((message, "mantatail"))
 
 
+def no_such_server(user: server.UserConnection, server_name: str) -> None:
+    """Sent when a user provides a non-existing server as an argument in a commant."""
```
Akuli, line 89:
```suggestion
    """Sent when a user provides a non-existing server as an argument in a command."""
```

tests/test_basic_functionality.py
```diff
@@ -252,6 +274,18 @@ def test_message_starting_with_colon(user_alice, user_bob, helpers):
...
 
 
+def test_whois_reply(user_alice, user_bob, helpers):
+    user_alice.sendall(b"WHOIS Alice\r\n")
+    assert helpers.receive_line(user_alice) == b":mantatail 311 Alice Alice AliceUsr 127.0.0.1 * :Alice's real name\r\n"
+    assert helpers.receive_line(user_alice) == b":mantatail 312 Alice Alice Mantatail :Running locally\r\n"
+    assert helpers.receive_line(user_alice) == b":mantatail 318 Alice Alice :End of /WHOIS list.\r\n"
+
+    user_alice.sendall(b"WHOIS Bob\r\n")
+    assert helpers.receive_line(user_alice) == b":mantatail 311 Bob Bob BobUsr 127.0.0.1 * :Bob's real name\r\n"
```
Akuli, line 284: Should the nickname really be repeated twice, even when you run WHOIS for someone else than yourself?

ThePhilgrim: I missed that I had already written this test several days ago. The other test would be the correct one. Removed this now.

